PR #2, opened 2021-06-20
Flesh out shopping list model functionality and API endpoints

danascheider:
## Context

Users need to be able to manage their shopping lists through the API. This PR adds resource routes for shopping lists.

### General

* Change response code from verifications controller to 204 instead of 201 since the resource may have been created previously

### Shopping List Model

* Add `master` column to database to establish concept of master list
* Add `title` column to database so users can differentiate lists
* Add validations:
  * One and only one master list per user
  * Uniqueness of title scoped to user
* Add callbacks:
  * Set default title on master lists or lists where no title has been specified
  * Create master list after creation of regular list if no master list exists yet
  * Prevent master list from being destroyed unless it is the user's only shopping list
  * Destroy master list automatically when the user destroys their last regular shopping list
* Overwrite `#to_json` method to include shopping list item children unless otherwise specified

### Shopping List Routes

* Add RESTful resource routes for shopping lists (index/create/show/update/destroy)

### Shopping List Item Model

* Add validations:
  * Validate description presence
  * Validate that quantity is present and is an integer greater than 0
* Add callbacks:
  * Before save, call `humanize` on description value to normalise and facilitate combining similar values on master lists
  * Abort save if description is being changed
  * Update master list items/quantities when list item is created, updated, or destroyed
* Add delegator so `:user` can be called directly on a list item

### User Model

* Add `#master_shopping_list` method to avoid having to use `shopping_lists.find_by(master: true)`

### Docs and Testing

* Add detailed API documentation
* Update factories for shopping lists
* Add and update tests

### DIFF
--- a/app/controllers/shopping_lists_controller.rb
+++ b/app/controllers/shopping_lists_controller.rb
@@ -9,7 +9,7 @@ class ShoppingListsController < ApplicationController
     shopping_list = current_user.shopping_lists.new
 
     if shopping_list.save
-      render json: shopping_list, status: :created
+      render json: { shopping_list: shopping_list, master_list: current_user.master_shopping_list }, status: :created
     else
       head :unprocessable_entity
     end
--- a/app/controllers/shopping_lists_controller.rb
+++ b/app/controllers/shopping_lists_controller.rb
@@ -2,7 +2,7 @@
 
 class ShoppingListsController < ApplicationController
   def index
-    render json: current_user.shopping_lists, status: :ok
+    render json: current_user.shopping_lists.includes(:shopping_list_items), status: :ok
   end
 
   def create
@@ -18,7 +18,7 @@ class ShoppingListsController < ApplicationController
   def show
     shopping_list = current_user.shopping_lists.find(params[:id])
 
-    render json: shopping_list, status: :ok
+    render json: shopping_list.to_json(include: :shopping_list_items), status: :ok
   rescue ActiveRecord::RecordNotFound
     head :not_found
   end
--- a/app/controllers/shopping_lists_controller.rb
+++ b/app/controllers/shopping_lists_controller.rb
@@ -26,7 +26,7 @@ class ShoppingListsController < ApplicationController
   def destroy
     shopping_list = current_user.shopping_lists.find(params[:id])
     shopping_list.destroy!
-    head :no_content
+    render json: { master_list: current_user.master_shopping_list }, status: :ok
   rescue ActiveRecord::RecordNotFound
     head :not_found
   end
--- a/app/controllers/shopping_lists_controller.rb
+++ b/app/controllers/shopping_lists_controller.rb
@@ -16,9 +16,9 @@ class ShoppingListsController < ApplicationController
   end
 
   def show
-    shopping_list = current_user.shopping_lists.find(params[:id])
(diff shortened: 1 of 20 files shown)


app/controllers/verifications_controller.rb
```diff
@@ -8,6 +8,6 @@ class VerificationsController < ApplicationController
   # token has been verified server-side and that there is
   # a corresponding user in the system.
   def verify_token
-    head :created
+    head :no_content
```
danascheider, line 11: `:created` might not be accurate since it's possible it was created on an earlier sign-in.

app/controllers/shopping_lists_controller.rb
```diff
@@ -23,11 +23,47 @@ def show
...
+  rescue ActiveRecord::RecordNotFound
+    head :not_found
+  end
+
   def destroy
     shopping_list = current_user.shopping_lists.find(params[:id])
-    shopping_list.destroy!
-    head :no_content
+
+    if shopping_list.master
```
danascheider, line 41: This will probably be able to be refactored out when I do [this card](https://trello.com/c/tPeoZBxn/3-return-error-for-non-get-methods-on-master-list-resource) since we'll be significantly restricting ability to work with master lists through the API.

app/models/shopping_list_item.rb
```diff
@@ -2,4 +2,74 @@
...
+
+    if item_on_master_list.quantity == quantity
+      item_on_master_list.destroy!
+    else
+      item_on_master_list.update!(quantity: item_on_master_list.quantity - quantity)
+    end
+  end
+
+  def prevent_changed_description
+    return true if new_record?
```
danascheider, line 60: Change to `before_update` callback and we wouldn't need this.

app/controllers/shopping_lists_controller.rb
```diff
@@ -23,11 +23,47 @@ def show
     head :not_found
   end
 
+  def update
+    shopping_list = current_user.shopping_lists.find(params[:id])
```
danascheider, line 27: Master lists should not be updateable via this route but that work will be done in [this card](https://trello.com/c/tPeoZBxn/3-return-error-for-non-get-methods-on-master-list-resource).

app/models/shopping_list.rb
```diff
@@ -3,4 +3,56 @@
...
+  validate :one_master_list_per_user
+  validates :title, uniqueness: { scope: :user_id }
+
+  before_create :set_default_title, if: :master_or_title_blank?
+  after_create :ensure_master_list_present
+  before_destroy :ensure_not_master, if: :other_lists_present?
+  after_destroy :destroy_master_list, unless: :other_lists_present?
+
+  def to_json(opts = {})
+    opts.merge!({ include: :shopping_list_items }) unless opts.has_key?(:include)
```
danascheider, line 16: This way the behaviour can still be changed by passing in `include: nil`.

spec/requests/shopping_lists_spec.rb
```diff
@@ -41,7 +41,7 @@
 
         it 'returns the new list' do
           create_shopping_list
-          expect(response.body).to eq ShoppingList.last.to_json
```
danascheider, line 44: Turns out that this will be the master list that gets created in the after_create callback.

PR #4, opened 2021-06-20
Fix bug where notes not getting updated on master list items when regular list items changed

danascheider:
## Context

[**Combine notes fields from regular shopping items on master list**](https://trello.com/c/4yLo85Ub/2-combine-notes-fields-from-regular-shopping-list-items-on-master-list)

We were losing notes on regular list items because they weren't being added to master list notes when added to the master list. We needed to change this behaviour:

### When a New Regular List Item Is Created

The master list should add that item's "notes" value, if any, to its own, separated from existing values with ` -- `.

### When a Regular Item Is Updated

The logic goes like this:
```
if regular item notes are unchanged
  do nothing
elsif regular item notes are changed from nil to a value
  concatenate the value to the master list item notes value, separated by ` -- `
elsif regular item notes are changed from a value to nil/empty
  remove that value from the master list item, also removing any leading or trailing ` -- `
elsif regular item notes are changed from one value to another
  change that value in the master list item notes as well, taking care to only replace the first occurrence
  (since the master list could have multiple note values that are the same associated with different items)
```

### When a Regular Item Is Destroyed

The logic goes like this:
```
if regular_item.notes.blank?
  do nothing
elsif master_list_item.notes == regular_list_item.notes
  set master_list_item notes to nil
else
  remove first instance of regular_list_item.notes value from master_list_item.notes, along with
  any leading or trailing ` -- `
```

### Considerations

No `notes` value should ever begin or end with ` -- `. It is possible users would add this value to the beginning or end of their notes and that could throw a wrench in things but such is life.

## Changes

* Lots of tests
* Modify `ShoppingListItem` code to effect changes explained under 'Context'

### DIFF
--- a/app/models/shopping_list_item.rb
+++ b/app/models/shopping_list_item.rb
@@ -60,7 +60,7 @@ class ShoppingListItem < ApplicationRecord
       item_on_master_list.destroy!
     else
       if notes.present?
-        new_notes = item_on_master_list.notes.gsub(/#{notes}/, '').gsub(/^ ?\-\- /, '').gsub(/ \-\- ?$/, '')
+        new_notes = item_on_master_list.notes.sub(/#{notes}/, '').gsub(/^ ?\-\- /, '').gsub(/ \-\- ?$/, '')
       end
 
       new_notes = nil unless defined?(new_notes) && new_notes.present?
--- a/app/models/shopping_list_item.rb
+++ b/app/models/shopping_list_item.rb
@@ -35,7 +35,7 @@ class ShoppingListItem < ApplicationRecord
   private
 
   def add_to_master_list
-    new_attrs = self.attributes.reject { |key, value| ['shopping_list_id', :shopping_list_id, 'id', :id].include?(key) }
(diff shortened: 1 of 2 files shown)


app/models/shopping_list_item.rb
```diff
@@ -15,9 +15,9 @@ class ShoppingListItem < ApplicationRecord
   delegate :user, to: :shopping_list
 
   def self.create_or_combine!(attrs)
-    list = attrs[:shopping_list] || ShoppingList.find(attrs[:shopping_list_id])
+    shopping_list = attrs[:shopping_list] || ShoppingList.find(attrs[:shopping_list_id])
```
danascheider, line 18: Calling this `list` was creating some issues in `byebug` and I wondered if that could be because of a bug so I changed the variable name. Specifically, placing a `byebug` below this line and then typing `list` in the repl led to it stepping through some other code instead of returning a value. Maybe `list` is a method defined by ActiveRecord or something, I didn't bother to look into it because changing the variable name fixed it regardless.

PR #17, opened 2021-07-05
Massive refactor moving to controller services and MasterListable concern

danascheider:
## Context

[**Routes for managing shopping list items**](https://trello.com/c/K4501FRt/18-routes-for-managing-shopping-list-items)
[**Use controller services to handle complex shopping list logic**](https://trello.com/c/1gjVL1ix/62-use-controller-services-to-handle-complex-shopping-list-logic)
[**Consider shopping list updated when items added, removed, or edited**](https://trello.com/c/5DEczwm5/56-consider-shopping-list-updated-when-items-added-removed-or-edited)

It started that I was just going to add some routes to allow users to manage shopping list items from the UI. It ended in a sea of carnage enveloping three different cards. As I got started, I realised that the refactoring of shopping list logic could not wait and that the refactor would only get uglier if I waited until the routes were done. So I went ahead with a planned refactor to use controller services to handle requests and also added a `MasterListable` concern adding validations, scopes, hooks, and methods to the `ShoppingList` model.

As cool as it was, I decided that the self-updating logic of shopping list items was too magical. Now, to update a master list, call one of the methods provided by `MasterListable`.

## Changes

* Create `MasterListable` concern module (with docs) and add it to `ShoppingList`
* Implement controller service objects to handle requests
* Change `shopping_list.shopping_list_items` and `list_item.shopping_list` to `shopping_list.list_items` and `list_item.list`
* Standardise API responses to always either include a resource or a JSON object with `"errors"` array

### Required Changes

## Considerations

Since there are going to be other list models with master list behaviour (such as an inventory list), I decided it would be good to separate some of the master list logic into its own module. An ActiveSupport concern was a good way to do this since it had built-in support for scopes, validations, and other ActiveRecord stuff.

Controller services are a pattern I've seen in the wild that can be very useful to abstract away overly complex controller logic.

I didn't like that lists are no longer self-managing, however, I was concerned that self-managing lists were going to be buggy and overly complex due to their magic. Now, instead of automagically updating the master list when a list item is changed/added/removed, you have to call one of the new `MasterListable` methods on the master list to update it:

```ruby
item = ShoppingListItem.combine_or_create!(params)
master_list.add_item_from_child_list(item)
```
The `MasterListable` methods handle logic for combining, creating, updating, and removing items from master lists.

### DIFF
--- a/app/controller_services/shopping_lists_controller/create_service.rb
+++ b/app/controller_services/shopping_lists_controller/create_service.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'service/created_result'
+require 'service/unprocessable_entity_result'
+
+class ShoppingListsController < ApplicationController
+  class CreateService
+    MASTER_LIST_ERROR = 'Cannot manually create a master shopping list'
+
+    def initialize(user, params)
+      @user = user
+      @params = params
+    end
+
+    def perform
+      return Service::UnprocessableEntityResult.new(errors: [MASTER_LIST_ERROR]) if params[:master]
+
+      shopping_list = user.shopping_lists.new(params)
+
+      if shopping_list.save
+        resource = new_master_list&.save ? [user.master_shopping_list, shopping_list] : shopping_list
+        Service::CreatedResult.new(resource: resource)
+      else
+        Service::UnprocessableEntityResult.new(errors: assemble_error_messages(shopping_list.errors))
+      end
+    end
+
+    private
+
+    attr_reader :user, :params
+
+    def assemble_error_messages(data)
+      data.map { |error| "#{error.attribute.capitalize} #{error.message}" }
+    end
+
+    def new_master_list
+      return user.shopping_lists.new(master: true, title: 'Master') if user.master_shopping_list.nil?
+    end
+  end
+end
--- a/app/controller_services/shopping_lists_controller/create_service.rb
+++ b/app/controller_services/shopping_lists_controller/create_service.rb
@@ -23,17 +23,13 @@ class ShoppingListsController < ApplicationController
         resource = preexisting_master_list ? shopping_list : [user.master_shopping_list, shopping_list]
         Service::CreatedResult.new(resource: resource)
       else
-        Service::UnprocessableEntityResult.new(errors: assemble_error_messages(shopping_list.errors))
+        Service::UnprocessableEntityResult.new(errors: shopping_list.error_array)
       end
     end
 
     private
 
     attr_reader :user, :params
-
-    def assemble_error_messages(data)
-      data.map { |error| "#{error.attribute.capitalize} #{error.message}" }
-    end
 
     def new_master_list
       return user.shopping_lists.new(master: true, title: 'Master') if user.master_shopping_list.nil?
--- a/app/controller_services/shopping_lists_controller/create_service.rb
+++ b/app/controller_services/shopping_lists_controller/create_service.rb
@@ -30,9 +30,5 @@ class ShoppingListsController < ApplicationController
     private
 
     attr_reader :user, :params
-
-    def new_master_list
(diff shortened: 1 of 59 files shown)


app/controllers/shopping_list_items_controller.rb
```diff
@@ -24,19 +17,10 @@ def list_item_params
     params.require(:shopping_list_item).permit(
       :description,
       :quantity,
-      :notes,
-      :shopping_list_id
+      :notes
```
danascheider, line 20: Currently we don't need to pass the list ID in so I'm taking it out of the params so it doesn't confuse things.

app/controller_services/shopping_lists_controller/create_service.rb
```diff
@@ -0,0 +1,42 @@
...
+    def initialize(user, params)
+      @user = user
+      @params = params
+    end
+
+    def perform
+      return Service::UnprocessableEntityResult.new(errors: [MASTER_LIST_ERROR]) if params[:master]
+
+      shopping_list = user.shopping_lists.new(params)
+      preexisting_master_list = user.master_shopping_list
```
danascheider, line 19: Check if the master shopping list exists already - it will be returned if so.

app/models/concerns/master_listable.rb
```diff
@@ -0,0 +1,181 @@
...
+  def create_master_list
+    self.class.find_or_create_by!(user: user, master: true)
+  end
+
+  def set_title_to_master
+    self.title = 'Master'
+  end
+
+  def abort_if_master_changed
+    throw :abort if master_changed? && !new_record?
```
danascheider, line 129: It'll also be marked as changed when the record is new so have to check that it isn't the initial create.

app/models/shopping_list_item.rb
```diff
@@ -1,38 +1,39 @@
 # frozen_string_literal: true
 
 class ShoppingListItem < ApplicationRecord
-  belongs_to :shopping_list
+  belongs_to :list, class_name: 'ShoppingList'
```
danascheider, line 4: Thank god we can stop it with the `shopping_list_item.shopping_list` and the `shopping_list.shopping_list_items` stuff now.

app/models/shopping_list_item.rb
```diff
@@ -1,38 +1,39 @@
...
-  delegate :user, to: :shopping_list
+  delegate :user, to: :list
 
   scope :index_order, -> { order(updated_at: :desc) }
 
   def self.combine_or_create!(attrs)
-    combine_or_new(attrs).save!
+    obj = combine_or_new(attrs)
+    obj.save!
+    obj
```
danascheider, line 20: This was returning `true` before.

docs/api/resources/shopping-lists.md
```diff
@@ -123,61 +125,6 @@ For a user with multiple lists:
 ]
 ```
 
-## GET /shopping_lists/:id
```
danascheider, line 126: We weren't using this route on the front end at all and don't really have plans to, so I decided to let this one go.

lib/controller/response.rb
```diff
@@ -0,0 +1,25 @@
...
+    def initialize(controller, result, options = {})
+      @controller = controller
+      @result = result
+      @options = options
+    end
+
+    def execute
+      if result.errors.blank? && result.resource.blank?
+        controller.head result.status
+      elsif result.errors.any?
```
danascheider, line 14: Decided to be conservative and return errors if there are any instead of a resource.

spec/factories/shopping_list_items.rb
```diff
@@ -2,9 +2,9 @@
 
 FactoryBot.define do
   factory :shopping_list_item do
-    shopping_list
+    association :list, factory: :shopping_list
 
-    description { 'Necklace' }
```
danascheider, line 7: Had to change this after adding the validation that description had to be unique for items on a given list.

PR #18, opened 2021-07-06
Add update route for shopping list items

danascheider:
## Context

[**Routes for managing shopping list items**](https://trello.com/c/K4501FRt/18-routes-for-managing-shopping-list-items)
[**Consider shopping list updated when items added, removed, or edited**](https://trello.com/c/5DEczwm5/56-consider-shopping-list-updated-when-items-added-removed-or-edited)

This is the second PR introducing routes for managing shopping list items. (The first one is [here](https://github.com/danascheider/skyrim_inventory_management/pull/16).) The first one introduced a `create` route, and this one adds an `update` route.

## Changes

* Create new `ShoppingListItemsController::UpdateService` class to handle the update action in the controller
* Add `:update` method to the `ShoppingListItemsController` that uses the new service class
* Add specs for update route (unit specs for the new service class as well as request specs)
* Bug fixes uncovered by the specs
* Update docs

### Required Changes

## Considerations

I continued the pattern introduced in #17 to move to controller services to handle controller logic.

### DIFF
--- a/app/controller_services/shopping_lists_controller/create_service.rb
+++ b/app/controller_services/shopping_lists_controller/create_service.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'service/created_result'
-require 'service/not_found_result'
 require 'service/unprocessable_entity_result'
 require 'service/method_not_allowed_result'
 require 'service/ok_result'
--- a/spec/controller_services/shopping_list_items_controller/update_service_spec.rb
+++ b/spec/controller_services/shopping_list_items_controller/update_service_spec.rb
@@ -52,6 +52,10 @@ RSpec.describe ShoppingListItemsController::UpdateService do
       it 'updates the updated_at timestamp on the list' do
         Timecop.freeze do
           perform
+          # This is another case of a rounding error in the CI environment. The server where
+          # GitHub Actions run seems to truncate the last few digits of the timestamp, resulting
+          # in things being not quite equal in that environment. Since it's not that important
+          # that it be that exact, I'm just using the `be_within` matcher.
           expect(shopping_list.reload.updated_at).to be_within(0.05.seconds).of(Time.now.utc)
         end
       end
--- a/spec/controller_services/shopping_list_items_controller/update_service_spec.rb
+++ b/spec/controller_services/shopping_list_items_controller/update_service_spec.rb
@@ -1,0 +1,145 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'service/ok_result'
+require 'service/no_content_result'
+require 'service/not_found_result'
+require 'service/method_not_allowed_result'
+require 'service/unprocessable_entity_result'
+
+RSpec.describe ShoppingListItemsController::UpdateService do
+  describe '#perform' do
+    subject(:perform) { described_class.new(user, list_item.id, params).perform }
+
+    let(:user) { create(:user) }
+    let(:shopping_list) { user.shopping_lists.create! }
+    
+    context 'when all goes well' do
+      let!(:list_item) { create(:shopping_list_item, list: shopping_list, quantity: 2) }
+      let(:params) { { quantity: 3 } }
+      let(:master_list) { user.master_shopping_list }
+      let(:scope) { ShoppingListItem.belonging_to_user(user) }
+
+      before do
+        master_list.add_item_from_child_list(list_item)
+      end
+
+      it 'updates the shopping list item', :aggregate_failures do
+        perform
+        expect(list_item.reload.quantity).to eq 3
+        expect(list_item.reload.notes).to be nil
+      end
+
+      it 'updates the item on the master list' do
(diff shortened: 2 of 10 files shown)


app/controller_services/shopping_lists_controller/create_service.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'service/created_result'
-require 'service/not_found_result'
```
danascheider, line 4: This wasn't actually being used in here.

spec/requests/shopping_list_items_spec.rb
```diff
@@ -123,8 +123,8 @@
               # has frozen because Rails (Postgres?) sets the last three digits of
               # the timestamp to 0, which was breaking stuff in CI (but somehow not
               # in dev).
-              expect(shopping_list.reload.updated_at).to be_within(0.25.seconds).of(t)
-              expect(master_list.reload.updated_at).not_to be_within(0.25.seconds).of(t)
+              expect(shopping_list.reload.updated_at).to be_within(0.05.seconds).of(t)
```
danascheider, line 126: Narrowed down the tolerances here because we know the rounding error in CI is considerably less than 0.05s and we want to get as close as possible.

spec/controller_services/shopping_list_items_controller/update_service_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+      it 'updates the shopping list item', :aggregate_failures do
+        perform
+        expect(list_item.reload.quantity).to eq 3
+        expect(list_item.reload.notes).to be nil
+      end
+
+      it 'updates the item on the master list' do
+        # Put the mocks in here because I don't want this much mocking in the other specs
+        allow(ShoppingListItem).to receive(:belonging_to_user).with(user).and_return(scope)
```
danascheider, line 34: These mocks ended up being necessary in a couple places due to the fact that I didn't want to test the whole `MasterListable` methods (since these are unit tests and they're covered in other unit tests). Unfortunately, because there are DB calls made in the controller service to fetch these objects, they aren't the same objects in memory as the ones in the spec. So I had to make sure the ones in the spec were getting passed around instead.

spec/controller_services/shopping_list_items_controller/update_service_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      end
+
+      it 'updates the updated_at timestamp on the list' do
+        Timecop.freeze do
+          perform
+          # This is another case of a rounding error in the CI environment. The server where
+          # GitHub Actions run seems to truncate the last few digits of the timestamp, resulting
+          # in things being not quite equal in that environment. Since it's not that important
+          # that it be that exact, I'm just using the `be_within` matcher.
+          expect(shopping_list.reload.updated_at).to be_within(0.05.seconds).of(Time.now.utc)
```
danascheider, line 58: I could probably tighten the tolerance further since it's only the last few digits, but 0.05 is honestly good enough. Users are unlikely to be able to make requests from the UI fast enough for race conditions to come up as a result of the timestamps not being perfect.

spec/controller_services/shopping_list_items_controller/update_service_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+        end
+      end
+    end
+
+    context 'when the shopping list item is not found' do
```
danascheider, line 63: Thought about testing that the master list isn't updated here but if the list item doesn't exist the app can't tell what list it's on anyway so wouldn't be able to identify which master list to update in the first place.

spec/requests/shopping_list_items_spec.rb
```diff
@@ -38,7 +38,7 @@
       end
       
       context 'when all goes well' do
-        let(:params) { "{\"shopping_list_item\":{\"description\":\"Corundum ingot\",\"quantity\":5,\"notes\":\"To make locks\",\"list_id\":#{shopping_list.id}}}" }
+        let(:params) { { shopping_list_item: { description: 'Corundum ingot', quantity: 5, notes: 'To make locks' } } }
```
danascheider, line 41: I decided to change these strings into POROs and just call `to_json` on them in the `subject` block. That's a lot nicer than having those ugly JSON strings with interpolated data and escaped quotes.

spec/requests/shopping_list_items_spec.rb
```diff
@@ -224,4 +224,194 @@
...
+          end
+        end
+
+        it 'returns the master list item and the regular list item', :aggregate_failures do
+          update_item
+
+          # The serialization isn't as simple as .to_json and it makes it hard to determine if the JSON
+          # string is correct as some attributes are out of order and the timestamps are serialized
+          # differently. Here we grab the individual items and then we'll filter out the timestamps to
+          # verify them.
```
danascheider, line 302: The values of the timestamps, to the extent they matter, are covered in other specs.

PR #24, opened 2021-07-07
Compare list item descriptions case-insensitively throughout code base

danascheider:
## Context

In #22, I made sure that the `ShoppingListItem#combine_or_new` method compared item descriptions case-insensitively when determining whether to create a new item or combine the attributes with those of an existing one. However, elsewhere in the codebase, description comparisons remain case-sensitive.

## Changes

* Change all description comparisons to be case-insensitive
* Update tests to ensure misbehaviour will be caught before it reaches prod

### Required Changes

### DIFF
--- a/app/controller_services/shopping_list_items_controller/destroy_service.rb
+++ b/app/controller_services/shopping_list_items_controller/destroy_service.rb
@@ -33,10 +33,6 @@ class ShoppingListItemsController < ApplicationController
       @master_list ||= shopping_list_item.list.master_list
     end
 
-    def master_list_item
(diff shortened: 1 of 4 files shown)


app/controller_services/shopping_list_items_controller/destroy_service.rb
```diff
@@ -33,10 +33,6 @@ def master_list
       @master_list ||= shopping_list_item.list.master_list
     end
 
-    def master_list_item
```
danascheider, line 36: This method wasn't being used anywhere but did have the same name as a variable used elsewhere in the file.

PR #28, opened 2021-07-12
Change name of master shopping list to aggregate shopping list

danascheider:
## Context

[**Change name of master shopping list to main shopping list**](https://trello.com/c/kdN0AKtD/65-change-name-of-master-shopping-list-to-main-shopping-list)

For inclusivity reasons, it would be better not to use the term "master" list. This PR changes the name of "master" lists to "aggregate" lists.

## Changes

* Change `master` lists to `aggregate` lists
* Change automatic title to "All Items"
* Update docs and specs

### Required Changes

## Considerations

I considered changing the name to a "Main" or "Primary" list. "Aggregate" sounds a little too finance-schoolish to me, but it was also the most descriptive and accurate name I could think of. I decided on the name "All Items" for the aggregate lists because I think this will be easier for users to understand and more intuitive to them. We'll want to avoid exposing users to the concept of "aggregate lists" because it's not a very user-friendly term. Honestly I thought "master" was by far the best term for this reason, but if it isn't the best term for other people then it isn't the best term for me.


danascheider:
I've tested all the front- and back-end changes together in dev.

PR #31, opened 2021-07-15
Migrations for Game model

danascheider:
## Context

[**Create Game model**](https://trello.com/c/YuwzNEXY/89-create-game-model)

We're creating a `Game` model that will enable users to have multiple games, each of which will have its own shopping lists. This PR includes the relevant database migrations.

## Changes

* Create new `games` table with reference to `users`
* Add reference to `games` to the `shopping_lists` table

### Required Changes

## Considerations

I've decided to set a `NOT NULL` constraint on the new foreign key in `shopping_lists`. This will necessitate nuking the production database (well, the shopping lists and their items, anyway) in order to run the migration. I'm not really using the app yet so I'm OK with that.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_07_15_080837) do
+ActiveRecord::Schema.define(version: 2021_07_15_082312) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -10,11 +10,20 @@
...
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_07_12_200541) do
+ActiveRecord::Schema.define(version: 2021_07_15_082312) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "games", force: :cascade do |t|
+    t.bigint "user_id", null: false
```
danascheider, line 19: I noticed that some of our other foreign keys are just regular integers. Do we want them to be `bigint` as well?

PR #33, opened 2021-07-15
Create Game model

danascheider:
## Context

[**Create Game model**](https://trello.com/c/YuwzNEXY/89-create-game-model)

After adding the migrations to create the `games` table and add the foreign key to `ShoppingLists` in #31, it's time to create the game model. This PR creates the `Game` model and all its associations and makes all the model specs pass.

## Changes

* Create `Game` model that `belongs_to :user` and `has_many :shopping_lists`
* Add specs for the new model
* Fix broken model specs

### Required Changes

## Considerations

I've decided to merge this into the feature branch without making all the request and controller services specs pass. There is a card for reworking those routes and it didn't make sense to do that much work and touch that many files in this PR when there are other cards for that work.

### DIFF
--- a/app/models/game.rb
+++ b/app/models/game.rb
@@ -4,7 +4,7 @@ require 'titlecase'
 
 class Game < ApplicationRecord
   belongs_to :user
-  has_many :shopping_lists
+  has_many :shopping_lists, dependent: :destroy
 
   validates :name, uniqueness: { scope: :user_id },
                    format: {
@@ -13,6 +13,10 @@ class Game < ApplicationRecord
                    }
 
   before_save :format_name
+
+  def aggregate_shopping_list
+    shopping_lists.find_by(aggregate: true)
+  end
 
   private
 
--- a/app/models/game.rb
+++ b/app/models/game.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'titlecase'
+
+class Game < ApplicationRecord
+  belongs_to :user
+  has_many :shopping_lists
+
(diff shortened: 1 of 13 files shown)


app/models/game.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  before_save :format_name
+
+  def aggregate_shopping_list
+    shopping_lists.find_by(aggregate: true)
+  end
+
+  private
+
+  def format_name
```
danascheider, line 23: There is a bug in this method but it already exists in the `ShoppingList` model in the `format_title` method, so I've made a card to fix the issues at the same time before this feature is shipped.

app/models/shopping_list.rb
```diff
@@ -31,7 +31,7 @@ def format_title
     return if aggregate
 
     if title.blank?
-      highest_number = user.shopping_lists.where("title like '%My List%'").pluck(:title).map { |title| title.gsub('My List ', '').to_i }.max || 0
```
danascheider, line 34: The name `title` is already used so I'm switching this to `t`.

db/migrate/20210715085102_drop_user_id_from_shopping_lists.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class DropUserIdFromShoppingLists < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :shopping_lists, :user_id, null: false
```
danascheider, line 5: Oopsie.

PR #35, opened 2021-07-15
Scope shopping list create route to game

danascheider:
## Context

[**Scope shopping list routes to games**](https://trello.com/c/1Om5C5ek/95-scope-shopping-list-routes-to-games)

We've changed the `ShoppingList` model so it now belongs to the `Game` model, but not all the shopping list routes have been updated to reflect this fact. This PR tackles the `create` route.

## Changes

* Scope shopping list creation route to game
* Add 404 error response in case the game is not found or doesn't belong to the authenticated user

### Required Changes

## Considerations

The changes needed to scope the route to `games` was not that big. Mostly it was tests that needed updating. There were also some updates needed to the `ShoppingListsController::CreateService` class, which now takes a `game_id` param in addition to the user and the shopping list params and has additional error handling for the case where the game is not found or does not belong to the authenticated user.

### DIFF
--- a/spec/factories/games.rb
+++ b/spec/factories/games.rb
@@ -4,7 +4,7 @@ FactoryBot.define do
   factory :game do
     user
     
-    sequence(:name) { |n| "My Game #{n}" }
+    sequence(:name) { |n| "Skyrim Game #{n}" }
 
     factory :game_with_shopping_lists do
       transient do
--- a/spec/factories/shopping_lists.rb
+++ b/spec/factories/shopping_lists.rb
@@ -4,7 +4,7 @@ FactoryBot.define do
   factory :shopping_list do
     game
 
-    sequence(:title) { |n| "My List #{n}" }
+    sequence(:title) { |n| "Shopping List #{n}" }
 
     factory :aggregate_shopping_list do
       aggregate { true }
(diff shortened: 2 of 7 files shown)


spec/factories/shopping_lists.rb
```diff
@@ -4,7 +4,7 @@
   factory :shopping_list do
     game
 
-    sequence(:title) { |n| "My List #{n}" }
+    sequence(:title) { |n| "Shopping List #{n}" }
```
danascheider, line 7: I didn't want the default title in the factory to interfere with the model's default title behaviour.

spec/factories/games.rb
```diff
@@ -4,7 +4,7 @@
   factory :game do
     user
     
-    sequence(:name) { |n| "My Game #{n}" }
+    sequence(:name) { |n| "Skyrim Game #{n}" }
```
danascheider, line 7: I didn't want the default name in the factory to interfere with the model's default naming behaviour.

spec/requests/shopping_lists_spec.rb
```diff
@@ -47,22 +49,71 @@
...
           end
 
           it 'returns only the newly created list' do
             create_shopping_list
-            expect(response.body).to eq(user.shopping_lists.last.to_json)
+            expect(response.body).to eq(game.shopping_lists.last.to_json)
+          end
+        end
+
+        context 'when the request does not include a body' do
```
danascheider, line 64: The docs mention this behaviour so I thought it would be good to have tests for it as well.

PR #36, opened 2021-07-15
Scope shopping lists to games in update route

danascheider:
## Context

[**Scope shopping list routes to games**](https://trello.com/c/1Om5C5ek/95-scope-shopping-list-routes-to-games)

We've changed things so the `ShoppingList` model now belongs to the `Game` model and not the `User` model, but the routes and controller services have not been updated to reflect the change.

## Changes

* Fix `ShoppingListsController::UpdateService` to handle the fact that shopping lists now belong to games
* Add scope to `ShoppingList` model allowing shopping lists to be found by their user rather than just their game
* Add method to `User` model to retrieve all of a user's shopping lists
* Use `touch: true` on shopping list items/shopping lists relation to make sure shopping list updates when a list item is changed/added/removed without any action in the controller service (tests don't pass yet because we haven't made shopping list item stuff work with the `Game` model yet)

### Required Changes

## Considerations

There were minimal changes required here since the shallowly nested routes don't really bring the game model into the equation. The biggest change was creating the `User#shopping_lists` method to retrieve all of the shopping lists belonging to an individual user, enabling the controller to find a shopping list belonging to the authenticated user without knowing what game that shopping list belongs to.

### DIFF
--- a/app/controller_services/shopping_list_items_controller/create_service.rb
+++ b/app/controller_services/shopping_list_items_controller/create_service.rb
@@ -24,7 +24,6 @@ class ShoppingListItemsController < ApplicationController
 
       ActiveRecord::Base.transaction do
         item.save!
-        shopping_list.touch
 
         if preexisting_item.blank?
           aggregate_list_item = aggregate_list.add_item_from_child_list(item)
--- a/app/controller_services/shopping_list_items_controller/destroy_service.rb
+++ b/app/controller_services/shopping_list_items_controller/destroy_service.rb
@@ -22,7 +22,6 @@ class ShoppingListItemsController < ApplicationController
 
       ActiveRecord::Base.transaction do
         shopping_list_item.destroy!
-        shopping_list.touch
         aggregate_list_item = aggregate_list.remove_item_from_child_list(shopping_list_item.attributes)
       end
       
--- a/app/controller_services/shopping_list_items_controller/update_service.rb
+++ b/app/controller_services/shopping_list_items_controller/update_service.rb
@@ -25,7 +25,6 @@ class ShoppingListItemsController < ApplicationController
       aggregate_list_item = nil
       ActiveRecord::Base.transaction do
         list_item.update!(params)
-        shopping_list.touch
         
         aggregate_list_item = aggregate_list.update_item_from_child_list(list_item.description, delta_qty, old_notes, params[:notes])
       end
--- a/docs/api/resources/shopping-lists.md
+++ b/docs/api/resources/shopping-lists.md
@@ -21,7 +21,7 @@ Like other resources in SIM, shopping lists are scoped to the authenticated user
 * [`GET /games/:game_id/shopping_lists`](#get-gamesgame_idshopping_lists)
 * [`GET /shopping_lists/:id`](#get-shopping_listsid)
 * [`POST /games/:game_id/shopping_lists`](#post-gamesgame_idshopping_lists)
-* [`PUT|PATCH /shopping_lists/:id`](#patch-shopping_listsid)
+* [`PATCH|PUT /shopping_lists/:id`](#patchput-shopping_listsid)
 * [`DELETE /shopping_lists/:id`](#delete-shopping_listsid)
 
 ## GET /games/:game_id/shopping_lists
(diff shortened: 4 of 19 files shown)


app/controller_services/shopping_list_items_controller/create_service.rb
```diff
@@ -24,7 +24,6 @@ def perform
 
       ActiveRecord::Base.transaction do
         item.save!
-        shopping_list.touch
```
danascheider, line 27: Should no longer be needed but we'll double check at the point where we make these tests pass again.

app/controller_services/shopping_list_items_controller/destroy_service.rb
```diff
@@ -22,7 +22,6 @@ def perform
 
       ActiveRecord::Base.transaction do
         shopping_list_item.destroy!
-        shopping_list.touch
```
danascheider, line 25: Should no longer be needed but we'll double check at the point where we make these tests pass again.

app/controller_services/shopping_list_items_controller/update_service.rb
```diff
@@ -25,7 +25,6 @@ def perform
       aggregate_list_item = nil
       ActiveRecord::Base.transaction do
         list_item.update!(params)
-        shopping_list.touch
```
danascheider, line 28: Should no longer be needed but we'll double check at the point where we make these tests pass again.

docs/api/resources/shopping-lists.md
```diff
@@ -332,9 +332,9 @@ Content-Type: application/json
 
 #### Statuses
 
-* 422 Unprocessable Entity
-* 405 Method Not Allowed
 * 404 Not Found
+* 405 Method Not Allowed
```
danascheider, line 336: I just rearranged these to be in ascending order.

spec/controller_services/shopping_list_items_controller/create_service_spec.rb
```diff
@@ -46,6 +46,14 @@
           expect(aggregate_list).to have_received(:add_item_from_child_list).with(shopping_list.list_items.last)
         end
 
+        it 'updates the list model itself' do
+          t = Time.now + 3.days
+          Timecop.freeze(t) do
+            perform
+            expect(shopping_list.reload.updated_at).to be_within(0.005.seconds).of(t)
```
danascheider, line 53: Although Timecop is exact, differences with timestamps in CI, where the last four digits are truncated to 0, make it necessary to use `be_within` here instead of `eq`.

spec/controller_services/shopping_list_items_controller/destroy_service_spec.rb
```diff
@@ -40,17 +40,15 @@
...
           t = Time.now + 3.days
-
           Timecop.freeze(t) do
             perform
             # use `be_within` even though the time will be set to the time Timecop
             # has frozen because Rails (Postgres?) sets the last three digits of
             # the timestamp to 0, which was breaking stuff in CI (but somehow not
             # in dev).
-            expect(shopping_list.reload.updated_at).to be_within(0.05.seconds).of(t)
-            expect(aggregate_list.reload.updated_at).not_to be_within(0.05.seconds).of(t)
```
danascheider, line 53: The aggregate list should have its timestamp changed too but it isn't important because when the order of shopping lists matters the aggregate list is first regardless.

spec/controller_services/shopping_list_items_controller/destroy_service_spec.rb
```diff
@@ -86,17 +84,15 @@
...
           t = Time.now + 3.days
-
           Timecop.freeze(t) do
             perform
             # use `be_within` even though the time will be set to the time Timecop
             # has frozen because Rails (Postgres?) sets the last three digits of
             # the timestamp to 0, which was breaking stuff in CI (but somehow not
             # in dev).
             expect(shopping_list.reload.updated_at).to be_within(0.05.seconds).of(t)
-            expect(aggregate_list.reload.updated_at).not_to be_within(0.05.seconds).of(t)
```
danascheider, line 99: Again, this was not really important.

spec/controller_services/shopping_list_items_controller/update_service_spec.rb
```diff
@@ -104,6 +105,14 @@
         expect(aggregate_list).not_to have_received(:update_item_from_child_list).with(list_item.description, 1, nil, nil)
       end
 
+      it 'does not update the shopping list itself' do
+        t = Time.now + 3.days
+        Timecop.freeze(t) do
+          perform
+          expect(shopping_list.reload.updated_at).not_to be_within(71.hours).of(t)
```
danascheider, line 112: It should suffice to use `be_within(3.days)` but just in case.

PR #38, opened 2021-07-16
Update shopping list item model

danascheider:
## Context

[**Update shopping list item model and routes to account for new relation between shopping lists and games**](https://trello.com/c/yhSEVWui/97-update-shopping-list-item-model-and-routes-to-account-for-new-relation-between-shopping-lists-and-games)

We haven't yet updated the shopping list item model and its associated functionality to account for the fact that shopping lists now belong to games and not users. It turned out that the model itself needed minimal changes, primarily to make sure it's easy to access the user to which the list item belongs without chaining models.

## Changes

* Ensure that `#user` methods are available on shopping lists and shopping list items to retrieve the user that owns the models
* Make sure that user and game models know which shopping list items belong to them
* Update specs to test new methods and change some wording to clarify associations

### Required Changes

## Considerations

The `Game#shopping_list_items` method might not actually be needed for this card given that list items will be found by user in the controller services, but it'll probably be useful eventually so I added it anyway.

### DIFF
--- a/app/models/concerns/aggregatable.rb
+++ b/app/models/concerns/aggregatable.rb
@@ -42,9 +42,6 @@ module Aggregatable
 
     serialize :list_items, class_name: 'Array'
 
-    scope :aggregate_first, -> { order(aggregate: :desc) }
-    scope :includes_items, -> { includes(:list_items) }
-
     validate :one_aggregate_list_per_game,        if: :is_aggregate_list?
     validate :not_named_all_items,                unless: :is_aggregate_list?
     validate :ensure_aggregate_list_is_aggregate, unless: :is_aggregate_list?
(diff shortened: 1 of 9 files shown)


app/models/concerns/aggregatable.rb
```diff
@@ -56,6 +53,11 @@ class AggregateListError < StandardError; end
     before_save :set_title_to_all_items,     if: :is_aggregate_list?
     before_destroy :abort_if_aggregate,      if: :has_child_lists?
     after_destroy :destroy_aggregate_list,   unless: -> { is_aggregate_list? || aggregate_has_other_children? }
+
+    scope :aggregate_first, -> { order(aggregate: :desc) }
```
danascheider, line 57: I just moved these down because I like to do scopes and methods last.

PR #39, opened 2021-07-16
Update shopping list item create route

danascheider:
## Context

[**Update shopping list item model and routes to account for new relation between shopping lists and games**](https://trello.com/c/yhSEVWui/97-update-shopping-list-item-model-and-routes-to-account-for-new-relation-between-shopping-lists-and-games)

We're in the process of updating all the resource routes to account for changed associations between the `ShoppingList`, `Game`, and `User` models. This PR takes on the create route for shopping list items.

## Changes

* Update specs to account for changed associations between models
* Add additional test cases

### Required Changes

## Considerations

Because shopping list item routes don't interact directly with games, we only had to change the specs to make them pass.

### DIFF
--- a/spec/controller_services/shopping_list_items_controller/create_service_spec.rb
+++ b/spec/controller_services/shopping_list_items_controller/create_service_spec.rb
@@ -13,14 +13,17 @@ RSpec.describe ShoppingListItemsController::CreateService do
     subject(:perform) { described_class.new(user, shopping_list.id, params).perform }
 
     let(:user) { create(:user) }
+    let(:game) { create(:game, user: user) }
 
     context 'when all goes well' do
-      let!(:aggregate_list) { create(:aggregate_shopping_list, user: user) }
-      let!(:shopping_list) { create(:shopping_list, user: user, aggregate_list: aggregate_list) }
+      let!(:aggregate_list) { create(:aggregate_shopping_list, game: game) }
+      let!(:shopping_list) { create(:shopping_list, game: game, aggregate_list: aggregate_list) }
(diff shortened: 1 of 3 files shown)


spec/controller_services/shopping_list_items_controller/create_service_spec.rb
```diff
@@ -13,14 +13,17 @@
...
     context 'when all goes well' do
-      let!(:aggregate_list) { create(:aggregate_shopping_list, user: user) }
-      let!(:shopping_list) { create(:shopping_list, user: user, aggregate_list: aggregate_list) }
+      let!(:aggregate_list) { create(:aggregate_shopping_list, game: game) }
+      let!(:shopping_list) { create(:shopping_list, game: game, aggregate_list: aggregate_list) }
       let(:params) { { description: 'Necklace', quantity: 2, notes: 'Hello world' } }
 
       before do
-        allow(user.shopping_lists).to receive(:find).and_return(shopping_list)
+        user_lists = user.shopping_lists
```
danascheider, line 24: Like in another recent PR, this broke for some reason and adding this code fixed it.

spec/requests/shopping_list_items_spec.rb
```diff
@@ -169,7 +255,7 @@
 
         it 'does not return content' do
           create_item
-          expect(response.body).to be_empty
+          expect(response.body).to be_blank
```
danascheider, line 258: Come to find out, `empty?` isn't defined on `nil`, so this would blow up if the body were `nil` when really we'd be OK with that.

PR #47, opened 2021-07-17
Fix bug in default titles

danascheider:
## Context

[**Fix bug in default titles**](https://trello.com/c/RWYeeWGP/94-fix-bug-in-default-titles)

There was a (suspected) bug where default titles of shopping lists and names of games could be set to the wrong value. This bug didn't end up actually being a significant bug due to the fact that non-numeric strings will be cast to `0` and the minimum value is `1`.

The only case where this bug would be an issue is the case where there are no existing default names/titles but there is, for instance, a list titled "My List -3". This would lead the first default list to be named "My List -2" since the max exists but is not zero. It's hardly a likely bug, but would be a little weird if it were to come up, so I fixed it. There is a foreseeable issue where a similar problem could occur if there were no default games but one game was named "My Game 2.6", which would evaluate to `2`. However, `.` is not currently an allowed character in shopping list titles or game names, so for now this issue does not exist.

## Changes

* Add additional specs for game and shopping list models to check default titling/naming behaviour
* Tweak code so it works if a list is titled "My List <negative integer>" or a game is named "My Game <negative integer>"
* Allow additional characters (`","`, `"-"`, and `"'"`) in shopping list titles to bring them in line with allowed game names 

### Required Changes

### DIFF
--- a/spec/models/shopping_list_spec.rb
+++ b/spec/models/shopping_list_spec.rb
@@ -141,27 +141,27 @@ RSpec.describe ShoppingList, type: :model do
       end
 
       context 'allowed characters' do
-        it 'allows alphanumeric characters and spaces' do
-          list = build(:shopping_list, title: 'My List 1  ')
+        it 'allows alphanumeric characters, spaces, commas, apostrophes, and hyphens' do
+          list = build(:shopping_list, title: "aB 61 ,'-")
           expect(list).to be_valid
         end
 
         it "doesn't allow newlines", :aggregate_failures do
           list = build(:shopping_list, title: "My\nList 1  ")
           expect(list).not_to be_valid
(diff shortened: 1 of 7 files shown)


spec/models/shopping_list_spec.rb
```diff
@@ -141,27 +141,27 @@
       end
 
       context 'allowed characters' do
-        it 'allows alphanumeric characters and spaces' do
-          list = build(:shopping_list, title: 'My List 1  ')
+        it 'allows alphanumeric characters, spaces, commas, apostrophes, and hyphens' do
+          list = build(:shopping_list, title: "aB 61 ,'-")
```
danascheider, line 145: Use a title that includes all possible characters.

PR #49, opened 2021-07-17
Fix Rubocop Gemfile and layout violations

danascheider:
## Context

We're in the process of adding Rubocop, primarily for linting/style purposes. This is the first of a few PRs fixing Rubocop violations that already existed in the code.

## Changes

* Run `rails rubocop:auto_correct` with everything but Bundler and Layout cops commented out

### Required Changes

## Considerations

All the violations here were autocorrectable, so all I really did was run Rubocop (with only the subset of cops enabled).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -40,6 +40,7 @@ require:
   - rubocop-performance
 AllCops:
   Exclude:
+    - 'bin/**/*'
     - 'db/schema.rb'
     - 'vendor/bundle/**/*'
   DisabledByDefault: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -322,1346 +322,1346 @@ Layout/TrailingWhitespace:
   Enabled: true
   AllowInHeredoc: false
 
-###############
-### Linting ###
-###############
+# ###############
+# ### Linting ###
+# ###############
 
-Lint/AmbiguousAssignment:
-  Enabled: true
-Lint/AmbiguousBlockAssociation:
-  Enabled: true
-Lint/AmbiguousOperator:
-  Enabled: true
-Lint/AmbiguousRegexpLiteral:
-  Enabled: true
-Lint/AssignmentInCondition:
-  Enabled: true
-Lint/BigDecimalNew:
-  Enabled: true
-Lint/CircularArgumentReference:
-  Enabled: true
-Lint/ConstantDefinitionInBlock:
-  Enabled: true
-Lint/Debugger:
-  Enabled: true
-Lint/DeprecatedClassMethods:
-  Enabled: true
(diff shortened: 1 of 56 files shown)


.rubocop.yml
```diff
@@ -141,8 +141,6 @@ Layout/EmptyLinesAroundAccessModifier:
   Enabled: true
 Layout/EmptyLinesAroundArguments:
   Enabled: true
-Layout/EmptyLinesAroundArguments:
```
danascheider, line 144: This was a duplicate

.rubocop.yml
```diff
@@ -312,9 +310,6 @@ Layout/SpaceInsideReferenceBrackets:
   Enabled: true
   EnforcedStyle: no_space
   EnforcedStyleForEmptyBrackets: no_space
-Layout/SpaceInsideReferenceBrackets:
```
danascheider, line 315: This was a duplicate as well.

bin/bundle
```diff
@@ -24,15 +24,17 @@ m = Module.new do
   def cli_arg_version
     return unless invoked_as_script? # don't want to hijack other binstubs
     return unless "update".start_with?(ARGV.first || " ") # must be running `bundle update`
+
     bundler_version = nil
-    update_index = nil
+    update_index    = nil
```
danascheider, line 29: I'm going to disable it from looking in `bin` from now on.

spec/controller_services/shopping_list_items_controller/create_service_spec.rb
```diff
@@ -33,12 +33,13 @@
...
         it 'adds a list item to the given list' do
-          expect { perform }.to change(shopping_list.list_items, :count).from(0).to(1)
+          expect { perform }
+            .to change(shopping_list.list_items, :count).from(0).to(1)
         end
 
         it 'assigns the correct values' do
-          params_with_string_keys = {}
-          params.each { |key, value| params_with_string_keys[key.to_s] = value }
+          params_with_string_keys                                     = {}
```
danascheider, line 41: It probably shouldn't actually be like this but it's a bug in Rubocop and the last thing I want is to already have `rubocop:disable` comments all over my code already, so I'll live with it.

spec/models/game_spec.rb
```diff
@@ -162,7 +162,8 @@
     end
 
     it 'returns all list items belonging to the game' do
-      items = game.shopping_lists.map { |list| list.list_items.to_a }.flatten.sort
+      items = game.shopping_lists.map {|list| list.list_items.to_a }
+                .flatten.sort
```
danascheider, line 166: This is actually pretty ugly the way it doesn't align with anything, but I guess that's how that cop works :/

PR #51, opened 2021-07-17
Fixes for Rubocop Naming, Style and Security violations

danascheider:
## Context

The largest group of [Rubocop](https://github.com/rubocop/rubocop) cops is the `Style` group. This PR is the result of correcting violations in these cops as well as the `Security` and `Naming` groups, primarily using Rubocop's autocorrect feature.

## Changes

* Run style and security cops against existing codebase
* Remove `Naming/FileName` cop, which quickly proved to be more trouble than it was worth

### Required Changes

## Considerations

I ended up removing the `Naming/FileName` cop, even though I like the idea of it, because it kept saying that spec files, `routes.rb`, etc. should define classes with certain names. I could've excluded all those files from that cop but at that point it would've been barely better than `rubocop:disable` comments.

### DIFF
--- a/app/controller_services/application_controller/authorization_service.rb
+++ b/app/controller_services/application_controller/authorization_service.rb
@@ -26,7 +26,7 @@ class ApplicationController < ActionController::API
     rescue GoogleIDToken::CertificateError => e
       Rails.logger.error "Problem with OAuth certificate -- #{e.message}"
       Service::UnauthorizedResult.new(errors: ['Invalid OAuth certificate'])
-    rescue => e
+    rescue StandardError => e
       Rails.logger.error "Internal Server Error: #{e.message}"
       Service::InternalServerErrorResult.new(errors: [e.message])
     end
--- a/app/controllers/shopping_list_items_controller.rb
+++ b/app/controllers/shopping_list_items_controller.rb
@@ -27,7 +27,7 @@ class ShoppingListItemsController < ApplicationController
     params.require(:shopping_list_item).permit(
       :description,
       :quantity,
-      :notes
+      :notes,
     )
   end
 end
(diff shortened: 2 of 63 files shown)


app/controller_services/application_controller/authorization_service.rb
```diff
@@ -26,7 +26,7 @@ def perform
     rescue GoogleIDToken::CertificateError => e
       Rails.logger.error "Problem with OAuth certificate -- #{e.message}"
       Service::UnauthorizedResult.new(errors: ['Invalid OAuth certificate'])
-    rescue => e
+    rescue StandardError => e
```
danascheider, line 29: Nice one, Rubocop!

app/controllers/shopping_list_items_controller.rb
```diff
@@ -27,7 +27,7 @@ def list_item_params
     params.require(:shopping_list_item).permit(
       :description,
       :quantity,
-      :notes
+      :notes,
```
danascheider, line 30: I don't like this style but it's more annoying to add an item to the end of a list, forget to add a comma to what was previously the last element, and get an error.

app/models/concerns/aggregatable.rb
```diff
@@ -60,19 +60,17 @@ class AggregateListError < StandardError; end
...
   def remove_item_from_child_list(attrs)
-    raise AggregateListError, 'remove_item_from_child_list method only available on aggregate lists' unless is_aggregate_list?
+    raise AggregateListError.new('remove_item_from_child_list method only available on aggregate lists') unless aggregate_list?
 
     existing_item = list_items.find_by('description ILIKE ?', attrs['description'])
 
-    if existing_item.nil? || existing_item.quantity < attrs['quantity']
-      raise AggregateListError, 'item passed to remove_item_from_child_list method is not represented on the aggregate list'
-    end
+    raise AggregateListError.new('item passed to remove_item_from_child_list method is not represented on the aggregate list') if existing_item.nil? || existing_item.quantity < attrs['quantity']
```
danascheider, line 73: I can't decide if it's a good or bad thing that I didn't turn on the line length cop.

PR #53, opened 2021-07-17
Run rubocop-rspec and rubocop-performance and make them pass

danascheider:
## Context

This is the second-to-last PR for running the newly-installed Rubocop against the existing code base. This one runs the rubocop-rspec cops.

## Changes

* Run rubocop-rspec cops with autocorrect
* Correct things to make them pass

### Required Changes

## Considerations

I could not convince Rubocop that two of the RSpec examples were not duplicates, so I was forced to put in our first `rubocop:disable` comments, a fact I deeply resent.


spec/controller_services/games_controller/update_service_spec.rb
```diff
@@ -32,8 +32,8 @@
     context 'when the params are invalid' do
       let!(:game) { create(:game) }
```
danascheider, line 33: I don't know why Rubocop doesn't think this one needs to be aligned with the others but I feel like if I change it it'll break.

PR #60, opened 2021-08-08
Create InventoryList and InventoryListItem models

danascheider:
## Context

[**Create InventoryList and InventoryListItem models**](https://trello.com/c/2Tm6DBTU/136-create-inventorylist-and-inventorylistitem-models)

We're kicking off the [Inventory Lists epic](https://trello.com/c/D6opBEo2/130-inventory-lists). The first step is to create the new models needed: the inventory list and the inventory list item. These models are almost identical to the shopping list and shopping list item models, except that the `InventoryListItem` has a `unit_weight` column (which will be added to shopping list items as part of [this card](https://trello.com/c/S2gwTYIX/154-add-unitweight-field-to-shopping-list-items)).

## Changes

* Create Aggregatable `InventoryList` model with migration
* Create `InventoryListItem` model with migration
* Add factories for new models
* Add specs for new models
* Make changes to other models as needed to define associations and utility methods and manage the association lifecycle

### Required Changes

## Considerations

There's a lot of duplication in the specs for `ShoppingList`/`InventoryList` and `ShoppingListItem`/`InventoryListItem`. Although some of the logic tested comes from the `Aggregatable` concern (and additional logic will be extracted to the `Listable` concern when that concern [is created](https://trello.com/c/3zzaSnq7/155-create-listable-concern-for-aggregatable-list-items)), it's important to test the logic in the models that include the concerns since it is core logic for them and since there is no other good way to test ActiveModel concerns.

### DIFF
--- a/app/models/inventory_list.rb
+++ b/app/models/inventory_list.rb
@@ -1,6 +1,17 @@
 # frozen_string_literal: true
 
 class InventoryList < ApplicationRecord
+  # Titles have to be unique per game as described in the API docs. They also can only
+  # contain alphanumeric characters and spaces with no special characters or whitespace
+  # other than spaces. Leading or trailing whitespace is stripped anyway so the validation
+  # ignores any leading or trailing whitespace characters.
+  validates :title,
+            uniqueness: { scope: :game_id, message: 'must be unique per game', case_sensitive: false },
+            format:     {
+                          with:    /\A\s*[a-z0-9 \-',]*\s*\z/i,
+                          message: "can only contain alphanumeric characters, spaces, commas (,), hyphens (-), and apostrophes (')",
+                        }
+
   # This has to be defined before including AggregateListable because its `included` block
   # calls this method.
   def self.list_item_class_name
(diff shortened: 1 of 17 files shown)


app/models/inventory_list.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  include Aggregatable
+
+  scope :index_order, -> { includes_items.aggregate_first.order(updated_at: :desc) }
+  scope :belonging_to_user, ->(user) { joins(:game).where(games: { user_id: user.id }).order('inventory_lists.updated_at DESC') }
+
+  private
+
+  def format_title
```
danascheider, line 32: Since this identical method is also present in the `ShoppingList` model (and a similar one is present in the `Game` model, it might be good to extract it somewhere, but I wasn't sure where would be a good place. Perhaps a second "Defaultable" concern would be possible, but it would have to contend with the fact that the defaultable field might differ from model to model.

app/models/inventory_list_item.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class InventoryListItem < ApplicationRecord
```
danascheider, line 3: Basically all the contents of this class will eventually be extracted to the `Listable` concern since it's all duplicated between this and `ShoppingListItem`.

PR #63, opened 2021-08-09
Implement logic around list item unit weights

danascheider:
## Context

[**Incorporate unit_weight logic into Aggregatable and Listable concerns**](https://trello.com/c/GSVOgowG/156-incorporate-unitweight-logic-into-aggregatable-and-listable-concerns)

## Changes

* Update `Listable::combine_or_new` method to update unit weight if a new, non-`nil` unit weight is included with the attributes
* Update `Aggregatable#add_item_from_child_list` and `Aggregatable#update_item_from_child_list` methods to make sure unit weight is updated on all matching items
* Modify `ShoppingListItemsController::CreateService` and `ShoppingListItemsController::UpdateService` to return all changed list items, not just the directly created/updated item and the aggregate list item

### Required Changes

## Considerations

This PR contains breaking changes. I considered making the changes to the controller services happen in a future PR, but the more I thought about it the more I saw no good reason to wait. Not doing it now would mean that items were getting updated and not returned from the API in the meantime.

The way I've implemented this logic will preclude setting unit weight to `nil` if the item already has a unit weight. In other words, once an item has a unit weight, it can't be removed. I don't love this but I prefer it to having unit weights on aggregate list items and items on other lists changed to `nil` when someone creates a new item without assigning a `unit_weight`. The `update_item_from_child_list` method has no way of distinguishing between a value being explicitly set to `nil` versus a value just not being set. It might be possible to modify the method to take a `:set_nil_unit_weight` option that would be set to true if the `unit_weight` is explicitly defined in the params. However, using this would invariably require some mind reading on the front end as we'd need to decide when an empty value was "intended" and when it was just left empty. I don't see a good way of making this determination, so for now we'll stick with the idea that unit weights cannot be unset once they are set.

### DIFF
--- a/app/models/concerns/aggregatable.rb
+++ b/app/models/concerns/aggregatable.rb
@@ -68,7 +68,7 @@ module Aggregatable
     raise AggregateListError.new('add_item_from_child_list method only available on aggregate lists') unless aggregate_list?
 
     if item.unit_weight
-      other_items = child_lists.map(&:list_items)
+      other_items = child_lists.all.map(&:list_items)
       other_items.flatten!
 
       other_items.each do |list_item|
--- a/app/models/concerns/aggregatable.rb
+++ b/app/models/concerns/aggregatable.rb
@@ -67,6 +67,15 @@ module Aggregatable
   def add_item_from_child_list(item)
     raise AggregateListError.new('add_item_from_child_list method only available on aggregate lists') unless aggregate_list?
 
+    if item.unit_weight
+      other_items = child_lists.map(&:list_items)
+      other_items.flatten!
+
+      other_items.each do |list_item|
+        list_item.update!(unit_weight: item.unit_weight) if list_item.description.casecmp(item.description).zero?
+      end
+    end
+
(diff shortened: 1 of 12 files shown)


spec/models/inventory_list_spec.rb
```diff
@@ -83,7 +83,7 @@
     describe 'aggregate lists' do
       context 'when there are no aggregate lists' do
         let(:game)           { create(:game) }
-        let(:aggregate_list) { build(:aggregate_shopping_list, game: game) }
```
danascheider, line 86: Whoopsie

app/models/concerns/aggregatable.rb
```diff
@@ -67,6 +67,15 @@ class AggregateListError < StandardError; end
   def add_item_from_child_list(item)
     raise AggregateListError.new('add_item_from_child_list method only available on aggregate lists') unless aggregate_list?
 
+    if item.unit_weight
+      other_items = child_lists.map(&:list_items)
+      other_items.flatten!
```
danascheider, line 72: I wanted to do this in one line but rubocop-performance suggests that way is slower.

PR #70, opened 2021-08-12
Endpoint to delete inventory list items

danascheider:
## Context

[**DELETE /inventory_list_items/:id endpoint**](https://trello.com/c/t4KU1uqK/143-delete-inventorylistitems-id-endpoint)

The final REST endpoint we need for inventory lists is the endpoint to destroy an inventory list item.

## Changes

* `InventoryListItemsController::DestroyService` class to handle `DELETE` requests for inventory list items
* Wire up the `DestroyService` to the controller
* Add `DELETE /inventory_list_items/:id` route
* Add request specs and unit specs for the new controller service
* Update docs

### Required Changes

## Considerations

I decided not to test the `Aggregatable#remove_item_from_child_list` logic in these tests because that functionality is already tested in tests for both the `InventoryList` class and the `ShoppingList` class.

### DIFF
--- a/docs/api/resources/shopping-list-items.md
+++ b/docs/api/resources/shopping-list-items.md
@@ -329,7 +329,7 @@ Authorization: Bearer xxxxxxxxxxx
 
 #### Example Body
 
-The API will return a 204 response if the list item has been destroyed along with the corresponding item on the aggregate list. This responsee does not include a body. On the other hand, if the aggregate list item has been updated, it will be returned and the status code will be 200.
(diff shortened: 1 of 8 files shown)


docs/api/resources/shopping-list-items.md
```diff
@@ -363,17 +363,7 @@ A 405 error, which is returned if the specified shopping list item is on an aggr
 ```json
 {
   "errors": [
-    "Cannot manually delete an aggregate shopping list"
-  ]
-}
-```
-
-A 422 error, returned as a result of a validation error, includes whichever errors prevented the list item from being created:
```
danascheider, line 371: This was an obvious copy-paste job gone wrong.

PR #78, opened 2021-09-15
Create canonical property and property models

danascheider:
## Context

[**Create Property model scoped to Game**](https://trello.com/c/l2GGGksv/169-create-property-model-scoped-to-game)

Users should be able to create the properties they are able to own in a given game.

## Changes

* Create `CanonicalProperty` model, with migration, to represent the actual properties a character can own in-game
* Add Rake task to populate canonical properties in the database
* Create `Property` model that `belongs_to` a game and corresponds to a canonical property
* Add optional associations where inventory lists and shopping lists can `belong_to` a property 

### Required Changes

## Considerations

Initially I didn't think we'd need a `CanonicalProperty` model because there are few enough canonical properties that relevant data for validations could be kept in constants in memory. But I decided to follow the pattern we're using for other items that have canonical attributes in-game and create a canonical model with a Rake task to populate the database with the canonical models.


app/models/canonical_property.rb
```diff
@@ -0,0 +1,70 @@
...
+            inclusion:  { in: VALID_NAMES, message: "must be an ownable property in Skyrim, or the Arch-Mage's Quarters" },
+            uniqueness: { message: 'must be unique' }
+
+  validates :hold,
+            presence:   true,
+            inclusion:  { in: VALID_HOLDS, message: 'must be one of the nine Skyrim holds, or Solstheim' },
+            uniqueness: { message: 'must be unique' }
+
+  validates :city,
+            inclusion:  { in: VALID_CITIES, message: 'must be a Skyrim city in which an ownable property is located', allow_blank: true },
```
danascheider, line 57: Technically Raven Rock is not a "Skyrim city" since it is in Solstheim but I think from the standpoint of the error message I might just be being pedantic.

spec/models/property_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+        property.canonical_property_id = canonical_property.id
+        property.name                  = canonical_property.name
+        property.hold                  = canonical_property.hold
+        property.city                  = nil
+        property.validate
+        expect(property.errors[:base]).to eq ['property attributes must match attributes of a property that exists in Skyrim']
+      end
+    end
+
+    describe 'arcane enchanter availability' do
```
danascheider, line 115: There really should be an 'alchemy lab availability' example group, but there are no properties in Skyrim where an alchemy lab is not a possible amenity, so it's really not that relevant anyway.

PR #85, opened 2022-04-29
Create CanonicalJewelryItem model

danascheider:
## Context

[**Create and populate canonical apparel models**](https://trello.com/c/hD3Ff0BA/152-create-and-populate-canonical-apparel-models)

The last of the canonical apparel item models we'll be creating is the CanonicalJewelryItem; the accompanying CanonicalJewelryItemsEnchantment model is a join table enabling multiple enchantments on a single piece of jewelry. (This join table is intended to add actual enchantments to special pieces of jewellery that have them - it isn't intended to associate _potential_ enchantments.

## Changes

* Migration to create the `canonical_jewelry_items` table
* `CanonicalJewelryItem` model class
* Migration to create `canonical_jewelry_items_enchantments` join table, enabling multiple enchantments to be associated to pieces of jewellery
* `CanonicalJewelryItemsEnchantment` model corresponding to the join table
* Remove default `unit_weight` value from the CanonicalClothingItem model migration since it hasn't run in production yet
* Light refactoring of tests to test that validated models' errors objects `include` an error message rather than testing the value of the total array
* Add missing associations to `CanonicalMaterial` model

### Required Changes

### DIFF
--- a/app/models/canonical_armor.rb
+++ b/app/models/canonical_armor.rb
@@ -18,5 +18,5 @@ class CanonicalArmor < ApplicationRecord
                          in:      %w[head body hands feet shield],
                          message: 'must be "head", "body", "hands", "feet", or "shield"',
                        }
-  validates :unit_weight, numericality: { greater_than_or_equal_to: 0 }
+  validates :unit_weight, presence: true, numericality: { greater_than_or_equal_to: 0 }
 end
--- a/app/models/canonical_material.rb
+++ b/app/models/canonical_material.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class CanonicalMaterial < ApplicationRecord
+  has_many :smithable_armors, through: :canonical_armors_smithing_materials, source: :canonical_armor
+  has_many :temperable_armors, through: :canonical_armors_tempering_materials, source: :canonical_armor
+
   validates :name, presence: true, uniqueness: true
   validates :unit_weight, presence: true, numericality: { greater_than_or_equal_to: 0 }
 end
--- a/app/models/canonical_material.rb
+++ b/app/models/canonical_material.rb
@@ -2,5 +2,5 @@
 
 class CanonicalMaterial < ApplicationRecord
   validates :name, presence: true, uniqueness: true
-  validates :unit_weight, numericality: { greater_than_or_equal_to: 0 }
+  validates :unit_weight, presence: true, numericality: { greater_than_or_equal_to: 0 }
 end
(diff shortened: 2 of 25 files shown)


app/models/canonical_material.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class CanonicalMaterial < ApplicationRecord
+  has_many :smithable_armors, through: :canonical_armors_smithing_materials, source: :canonical_armor
+  has_many :temperable_armors, through: :canonical_armors_tempering_materials, source: :canonical_armor
```
danascheider, line 5: These statements should've been added when the model was first created.

app/models/canonical_armor.rb
```diff
@@ -18,5 +18,5 @@ class CanonicalArmor < ApplicationRecord
                          in:      %w[head body hands feet shield],
                          message: 'must be "head", "body", "hands", "feet", or "shield"',
                        }
-  validates :unit_weight, numericality: { greater_than_or_equal_to: 0 }
+  validates :unit_weight, presence: true, numericality: { greater_than_or_equal_to: 0 }
```
danascheider, line 21: This way if the unit weight is missing, the errors array will specifically indicate that the unit weight can't be blank.

app/models/canonical_material.rb
```diff
@@ -1,6 +1,9 @@
...
+
   validates :name, presence: true, uniqueness: true
-  validates :unit_weight, numericality: { greater_than_or_equal_to: 0 }
+  validates :unit_weight, presence: true, numericality: { greater_than_or_equal_to: 0 }
```
danascheider, line 8: Again, ensures a clear message to users in the event they leave unit weight blank that the change they need to make is to add a unit weight.

db/migrate/20220429080952_create_canonical_clothing_items.rb
```diff
@@ -4,7 +4,7 @@ class CreateCanonicalClothingItems < ActiveRecord::Migration[6.1]
   def change
     create_table :canonical_clothing_items do |t|
       t.string :name, null: false, unique: true
-      t.decimal :unit_weight, precision: 5, scale: 1, default: 1.0
```
danascheider, line 7: It was better to remove a default unit weight since unit weights vary enough there's no real reason to assume that would be the weight of any particular item.

spec/models/canonical_property_spec.rb
```diff
@@ -6,46 +6,57 @@
   subject(:property) { described_class.new }
 
   describe 'validations' do
-    it 'must have a valid name' do
-      property.validate
-      expect(property.errors[:name]).to eq ["can't be blank", "must be an ownable property in Skyrim, or the Arch-Mage's Quarters"]
-    end
+    describe 'name' do
```
danascheider, line 9: This makes it more consistent with elsewhere where validations are tested.

PR #77, opened 2021-08-28
Update object modelling feature branch

danascheider:
## Epic

[**Update Object Modelling**](https://trello.com/c/taLhKs0V/166-update-object-modelling)

## TODO

* [ ] Restart production server so changed initialiser works
* [ ] Run Rake task `canonical_models:sync:all` task in production (instructions in docs)
* [x] Remove feature branch from list of branches to run CI against

### [Change InventoryListItem name to InventoryItem](https://trello.com/c/o4Jw16EI/170-change-inventorylistitem-name-to-inventoryitem)

* [x] **[Change InventoryListItem name to InventoryItem](https://github.com/danascheider/skyrim_inventory_management/pull/76) [merged]:** Change name of `InventoryListItem` model to `InventoryItem` along with related classes and objects

### [Create Property model scoped to Game](https://trello.com/c/l2GGGksv/169-create-property-model-scoped-to-game)

* [x] **[Create canonical property and property models](https://github.com/danascheider/skyrim_inventory_management/pull/78) [merged]:** Create a `CanonicalProperty` and `Property` model to represent ownable properties in Skyrim

### [Create and populate canonical apparel models](https://trello.com/c/hD3Ff0BA/152-create-and-populate-canonical-apparel-models)

* [x] **[Create CanonicalArmor model](https://github.com/danascheider/skyrim_inventory_management/pull/80) [merged]:** Create a `CanonicalArmor` model to represent standard and unique armours that actually exist in Skyrim
* [x] **[Create many-to-many association between canonical armor and enchantments](https://github.com/danascheider/skyrim_inventory_management/pull/81) [merged]:** Enable canonical armour pieces to be associated to enchantments (those that are actually on quest items or special items, not all available for a specific armour type)
* [x] **[Create CanonicalMaterial model and join tables with CanonicalArmor](https://github.com/danascheider/skyrim_inventory_management/pull/82) [merged]:** Enable canonical armour models to be associated with the materials required to forge or improve them
* [x] **[Create CanonicalClothingItem model](https://github.com/danascheider/skyrim_inventory_management/pull/83) [merged]:** Create the `CanonicalClothingItem` model to represent mages' robes and plain clothes
* [x] **[Add body_slot field to CanonicalClothingItems](https://github.com/danascheider/skyrim_inventory_management/pull/84) [merged]:** Add missing field that will allow SIM to identify which enchantments are available for a particular clothing item
[...]

PR #86, opened 2022-05-03
Populate canonical apparel models from JSON data

danascheider:
## Context

* [**Create and populate canonical apparel models**](https://trello.com/c/hD3Ff0BA/152-create-and-populate-canonical-apparel-models)
* [**Associate canonical material models to models that use them**](https://trello.com/c/nvOHxGdS/153-associate-canonical-material-models-to-models-that-use-them)

We are populating the canonical models that will be used to validate user-generated data. This involves creating the models (#80, #81, #82, #83, #84, #85) and populating them with data. This card deals with the models representing apparel items - our largest category to date by far. The data are populated, like other canonical data have been, with Rake tasks using JSON data to create the objects in the database. This PR ended up encompassing a large number of changes to the database schema, models, and tests as data collection and entry revealed flaws in the original design. It also includes data validation and cleansing work.

## Changes

* Data for all canonical building and smithing materials that exist in the game
* Data for all canonical apparel items (clothing, jewellery, and armour) that exist in the game
* Rake task populating these models in the database and building their associations

### Required Changes

## Considerations

This PR represents a very iterative and non-linear process of finding the needed tables, columns, and associations. Because Skyrim's internal data model is exceedingly complex - indeed, this exact fact is what makes the game so interesting - it was difficult to find models and relationships that (more or less) accurately captured the aspects of the game that SIM is interested in without going too far into the weeds on details that are minimally relevant. Consequently, details of SIM's domain models, database schema, and associations were ironed out painstakingly as we learnt more about the nature of the objects in the game we are calling "canonical objects". It is for this reason that the reader will see commits updating the latest database migrations and model files with minor details, such as the addition or removal of a unique or not null constraint. In some cases (such as the first commit in this PR) these details were changed back in later commits. The real challenge here is making our canonical models rigorous enough to provide meaningful validation for user-generated data while keeping them flexible enough to find the endless exceptions to rules in game.

In addressing these challenges, I have made a number of critical decisions, which I will explain in detail in this section.

### Process

Collection and entry of data constituted a majority of the work for this PR. Initially, I created some JSON files like the ones previously made in this epic and was entering data by hand. However, when I saw the truly immense volume and extreme complexity of the data I was dealing with, I decided that, rather than enter 1500 armour items into a JSON file in alphabetical order, it would be better to enter the items on a spreadsheet in [Google Sheets](https://docs.google.com/spreadsheets/d/1Vl3DasbrcbNwvuGSsrhzk6MByM4Q5WcQG0_sMijd380/edit#gid=0). This enabled some copying and pasting from the wikis and forums (like [the Elder Scrolls Wiki](https://elderscrolls.fandom.com/wiki/The_Elder_Scrolls_Wiki)) where I have gotten the item information. I then downloaded the spreadsheets as CSV files and merged them with the existing JSON files for the corresponding apparel types using Rake tasks. (I never committed these Rake tasks because they were so narrowly applicable to merging these specific data sets that they wouldn't even work on the same data sets in their final form.)

### Modifying Migration Files from Previous PRs

Firstly, the reader will notice that this PR modifies migration files created in the earlier linked PRs. Ordinarily these PRs would have been deployed and the migrations run in production. Had that been the case, I would've generated new migrations to make the changes. However, since those PRs had not yet been deployed (as they were merged into the [feature branch](https://github.com/danascheider/skyrim_inventory_management/pull/77)), I decided to simply roll back the database each time I needed to make a change and in that way reduce the number of migrations (which is already large).

### The `magical_effects` column

Sometimes, in Skyrim, a special item is enchanted with an effect other than one of the standard canonical enchantments. The possible effects are very unpredictable. For that reason, I added a text field, `magical_effects`, to each table (armour, jewellery, and clothing) for the purpose of describing these miscellaneous effects. This field does not prevent the item from having standard enchantments; indeed, many items do have both standard enchantments and various other magical effects.

### Names and Item Codes

Skyrim offers a unique code for every object in the game. Names, however, turn out not to be unique.Although SIM doesn't strictly need to have any awareness of Skyrim internals, since we needed a unique identifier for the items and we learnt that that could not be the name, it made sense to use them. This may come in useful in the future, and it was, again, easier to add the codes to the items now than to have to add them to each model in the database in the future.

Skyrim also frequently has several distinct items that differ solely by appearance in the game or even on internal implementation. These are frequently mundane items unlikely to be sought out or collected, such as "Clothes", of which there are several distinct variants. In these cases, when the items had no other distinct features of interest to SIM (same name, weight, enchantments, etc.), I chose one of the item IDs essentially at random (the first one listed on the given site) and added that single item to the data. Should SIM want to distinguish between these items in the future, it will be very possible to do so, as additional variants can be added at any time.


app/models/canonical_armor.rb
```diff
@@ -15,8 +29,8 @@ class CanonicalArmor < ApplicationRecord
   validates :body_slot,
             presence:  true,
             inclusion: {
-                         in:      %w[head body hands feet shield],
-                         message: 'must be "head", "body", "hands", "feet", or "shield"',
+                         in:      %w[head body hands feet hair shield],
```
danascheider, line 32: This turns out to be a relevant body slot for Skyrim, and it applies only to one or two specific items.

app/models/canonical_armor.rb
```diff
@@ -1,11 +1,25 @@
 # frozen_string_literal: true
 
 class CanonicalArmor < ApplicationRecord
-  has_many :enchantments, through: :canonical_armors_enchantments
-  has_many :smithing_materials, through: :canonical_armors_smithing_materials, source: :canonical_materials
-  has_many :tempering_materials, through: :canonical_armors_tempering_materials, source: :canonical_materials
+  has_many :canonical_armors_enchantments, dependent: :destroy
```
danascheider, line 4: This stuff was just wrong in the original PRs - I forgot you had to include this association in the model file.

app/models/canonical_armor.rb
```diff
@@ -1,11 +1,25 @@
...
+  has_many :enchantments,
+           -> { select 'enchantments.*, canonical_armors_enchantments.strength as enchantment_strength' },
```
danascheider, line 6: This enables us to call `item.enchantments.first.enchantment_strength` to get the `strength` value off the join table.

app/models/canonical_clothing_item.rb
```diff
@@ -1,11 +1,15 @@
...
+           -> { select 'enchantments.*, canonical_clothing_items_enchantments.strength as enchantment_strength' },
+           through: :canonical_clothing_items_enchantments
 
-  validates :name, presence: true, uniqueness: true
+  validates :name, presence: true
+  validates :item_code, presence: true, uniqueness: { message: 'must be unique' }
   validates :unit_weight, numericality: { greater_than_or_equal_to: 0 }
   validates :body_slot,
             presence:  true,
-            inclusion: { in: %w[head hands body feet shield], message: 'must be "head", "hands", "body", "feet", or "shield"' }
```
danascheider, line 10: Shields can't be a clothing item, they are always armour.

db/migrate/20220429040232_create_canonical_armors_enchantments.rb
```diff
@@ -6,6 +6,8 @@ def change
       t.references :canonical_armor, null: false, foreign_key: true
       t.references :enchantment, null: false, foreign_key: true
 
+      t.decimal :strength, scale: 2, precision: 5
```
danascheider, line 9: `Muffle` has a possible fractional strength value.

db/migrate/20220429031113_create_canonical_armors.rb
```diff
@@ -4,11 +4,17 @@ class CreateCanonicalArmors < ActiveRecord::Migration[6.1]
...
+      t.string :item_code, null: false, unique: true
       t.string :weight, null: false
       t.string :body_slot, null: false
-      t.decimal :unit_weight, precision: 5, scale: 1
+      t.string :magical_effects
+      t.decimal :unit_weight, precision: 5, scale: 2
       t.boolean :dragon_priest_mask, default: false
       t.boolean :quest_item, default: false
+      t.boolean :unique_item, default: false
+      t.boolean :enchantable, default: true
```
danascheider, line 15: Not all non-magical items can be enchanted. This should always be set to false for models that are canonically enchanted but is really intended to determine whether non-enchanted items can _be_ enchanted.

db/migrate/20220429042245_create_canonical_materials.rb
```diff
@@ -3,11 +3,12 @@
...
   def change
     create_table :canonical_materials do |t|
-      t.string :name, null: false, unique: true
+      t.string :name, null: false
+      t.string :item_code, null: false, unique: true
       t.boolean :building_material, default: false
       t.boolean :smithing_material, default: false
-      t.decimal :unit_weight, precision: 5, scale: 1
-      t.index :name, unique: true
+      t.decimal :unit_weight, precision: 5, scale: 2, null: false
```
danascheider, line 10: Turns out Skyrim internally tracks numbers as decimals with 2 places.

db/migrate/20220429060600_create_canonical_armors_smithing_materials.rb
```diff
@@ -11,7 +11,7 @@ def change
                    null:        false,
                    foreign_key: true,
                    index:       { name: :index_canonical_armors_smithing_mats_on_canonical_mat_id }
-      t.integer    :count,
+      t.integer    :quantity,
```
danascheider, line 14: The name 'quantity' seemed more appropriate than 'count', which is more generic and less self-explanatory.

db/migrate/20220429084831_add_body_slot_to_canonical_clothing_items.rb
```diff
@@ -1,7 +0,0 @@
-# frozen_string_literal: true
-
-class AddBodySlotToCanonicalClothingItems < ActiveRecord::Migration[6.1]
```
danascheider, line 3: I reconsidered this migration since it was easy to add the body_slot column to the original migration creating the table.

lib/tasks/populate_canonical_models.rake
```diff
@@ -88,12 +88,209 @@ namespace :canonical_models do
       end
     end
 
+    desc 'Populate or update canonical building and smithing materials from JSON data'
```
danascheider, line 91: These Rake tasks are kind of ugly but they're also [about to get significantly modified](https://trello.com/c/YdoBROVq/161-modify-rake-tasks-for-populating-canonical-items-to-clean-up-items-not-in-json-files), so I've decided to leave them as-is. I think they're reasonably readable so I'm deferring the refactor until I do the linked card.

PR #87, opened 2022-05-04
Clean up DB models not present in JSON data in Rake tasks

danascheider:
## Context

[**Modify Rake tasks for populating canonical items to clean up items not in JSON files**](https://trello.com/c/YdoBROVq/161-modify-rake-tasks-for-populating-canonical-items-to-clean-up-items-not-in-json-files)

The Rake tasks for populating canonical items currently add models not currently in the database and updates those that are present in both the database and the JSON data. However, they don't remove objects or associations present in the database but not the JSON data. We want the Rake tasks to enable a full _sync_ of the database with the JSON data.

## Changes

* Add `preserve_existing_records` arg to Rake tasks to allow developers to control whether existing records no longer present in JSON data should be removed from the database or not
* Remove records and associations not present in JSON data if `preserve_existing_records` is set to `false`
* Update association join models when existing associations found
* Normalise jewellery item data to always include `enchantments` and `materials` arrays, even if empty
* Normalise clothing item data to always include `enchantments` arrays, even if empty
* Use `item_code` as unique identifiers for materials since names may not be unique
* Update docs

### Required Changes

## Considerations

Although, in general, we'll want the Rake tasks to perform a sync of the database with the JSON data, it's possible a situation might arise where we want items to be added or updated in the database but not destroyed if they don't exist in the JSON data anymore. For this reason, I added the `preserve_existing_records` arg to the Rake tasks. Its default value is `false`, meaning that, by default, the tasks will do a full sync. However, if the developer running the task sets this arg to something other than `false`, the database will be updated non-destructively.

I debated whether to add this option because it's hard to imagine a case where we wouldn't want a full sync. However, it was simple enough to create a parameter to preserve the current behaviour, so it seemed reasonable to do that just in case.


lib/tasks/populate_canonical_models.rake
```diff
@@ -2,57 +2,80 @@
 
 require 'json'
 
+FALSEY_VALUES = [false, 'false'].freeze
```
danascheider, line 5: rubocop-performance indicates that including array literals in a loop unnecessarily slows down execution. Since this array occurs in multiple tasks, it made more sense to declare the constant outside the tasks so I didn't have to declare it more than once. Ordinarily, arguments passed into Rake tasks are typecast to booleans, but in case the command line argument `false` is interpreted as a string, I'm checking for the string value `'false'` as well.

lib/tasks/populate_canonical_models.rake
```diff
@@ -89,208 +119,270 @@ namespace :canonical_models do
...
-        rescue ActiveRecord::RecordInvalid => e
-          Rails.logger.error "Validation error saving canonical jewelry item \"#{data[:attributes][:item_code]}\": #{e.message}"
-          raise e
-        rescue StandardError => e
-          Rails.logger.error "Unknown error saving canonical jewelry item \"#{data[:attributes][:item_code]}\": #{e.message}"
-          raise e
-        end
-      end
-    end
+    task :canonical_jewelry,
```
danascheider, line 150: I was very conflicted about this line and the other lines like it. I feel it would've read better had I kept it all on one line, but that line was simply too long. I found it would be less readable if the code was all on one line, despite the multi-line way not being the best either.

lib/tasks/populate_canonical_models.rake
```diff
@@ -89,71 +119,97 @@ namespace :canonical_models do
...
-          Rails.logger.error "Unknown error saving canonical material \"#{canonical_material_attributes[:name]}\": #{e.message}"
+          Rails.logger.error "Unknown error saving canonical material \"#{canonical_material_attributes[:item_code]}\": #{e.message}"
           raise e
         end
       end
     end
 
     desc 'Populate or update canonical jewelry items from JSON data'
-    task canonical_jewelry: :environment do
+    # rubocop:disable Layout/BlockAlignment
```
danascheider, line 150: Ordinarily if a cop causes problems, we disable it in the `.rubocop.yml` file. This time, though, I was conflicted enough about my decision to break the block-starting line into multiple lines that I didn't want to expressly accommodate that kind of code style in our cops. Especially since it's only Rake tasks, I think it makes sense to only disable it here.

lib/tasks/populate_canonical_models.rake
```diff
@@ -89,71 +119,101 @@ namespace :canonical_models do
...
+            material = CanonicalMaterial.find_by(item_code: m[:item_code])
+
+            if material.present? && item.canonical_materials.exclude?(material)
+              CanonicalJewelryItemsCanonicalMaterial.create!(
+                canonical_jewelry_item: item,
+                canonical_material:     material,
+                quantity:               m[:quantity],
+              )
+            elsif item.canonical_materials.include?(material)
+              item.canonical_jewelry_items_canonical_materials
```
danascheider, line 194: Associations' join models should be updated if they already exist.

PR #88, opened 2022-05-04
Create CanonicalIngredient model

danascheider:
## Context

[**Create CanonicalIngredient model**](https://trello.com/c/rXYIM2S2/156-create-canonicalingredient-model)

We are creating canonical models as a way of running complex validations on user input. We are creating each model and then adding JSON data to populate it with actual in-game objects using a Rake task.

## Changes

* Migrations to create the `canonical_ingredients` and `alchemical_properties_canonical_ingredients` tables
* `CanonicalIngredient` model with association and validations
* Tests for validations
* Add unique indices to join tables to ensure each record represents a unique association

### Required Changes

## Considerations

I noticed in the course of this work that there was a unique composite index on one join table ensuring the uniqueness of its records. I decided that all join tables should either have or not have this index. Since we want all associations to be unique, I opted to add the index to the other join tables. It wasn't until after I completed this work that I really noticed it brought the number of changed files from 13 to 33 😅

### DIFF
--- a/app/models/canonical_armors_smithing_material.rb
+++ b/app/models/canonical_armors_smithing_material.rb
@@ -4,7 +4,6 @@ class CanonicalArmorsSmithingMaterial < ApplicationRecord
   belongs_to :canonical_armor
   belongs_to :canonical_material
 
-  validates :canonical_armor_id, presence: true
-  validates :canonical_material_id, presence: true
+  validates :canonical_armor_id, uniqueness: { scope: :canonical_material_id, message: 'must form a unique combination with canonical material' }
   validates :quantity, numericality: { only_integer: true, greater_than: 0 }
 end
--- a/db/migrate/20220504045733_create_alchemical_properties_canonical_ingredients.rb
+++ b/db/migrate/20220504045733_create_alchemical_properties_canonical_ingredients.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class CreateAlchemicalPropertiesCanonicalIngredients < ActiveRecord::Migration[6.1]
+  def change
+    create_table :alchemical_properties_canonical_ingredients do |t|
+      t.references :alchemical_property,
+                   null:        false,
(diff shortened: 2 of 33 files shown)


db/migrate/20220504045733_create_alchemical_properties_canonical_ingredients.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class CreateAlchemicalPropertiesCanonicalIngredients < ActiveRecord::Migration[6.1]
```
danascheider, line 3: I thought about creating this as part of the previous migration since it's so closely related to that model. However, doing it more atomically allows us to roll back smaller changes at a time, so I went with multiple migrations instead.

app/models/canonical_armors_smithing_material.rb
```diff
@@ -4,7 +4,6 @@ class CanonicalArmorsSmithingMaterial < ApplicationRecord
   belongs_to :canonical_armor
   belongs_to :canonical_material
 
-  validates :canonical_armor_id, presence: true
-  validates :canonical_material_id, presence: true
```
danascheider, line 8: Doing this really is redundant since the `belongs_to` association already takes care of it.

db/migrate/20220504081145_add_composite_indexes_to_join_tables.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class AddCompositeIndexesToJoinTables < ActiveRecord::Migration[6.1]
+  def change
+    add_index :canonical_armors_enchantments,
+              %i[canonical_armor_id enchantment_id],
+              unique: true,
+              name:   'index_can_armors_enchantments_on_can_armor_id_and_ench_id'
```
danascheider, line 8: I had to make these awful shortened index names because Postgres limits them to 63 characters.

PR #92, opened 2022-05-10
Create CanonicalWeapon model

danascheider:
## Context

[**Create CanonicalWeapon model**](https://trello.com/c/Y9xsgPVG/157-create-canonicalweapon-model)

Work continues to add canonical models to SIM. We've determined that the models will all have to be created up front. The next model is the `CanonicalWeapon` model, representing both unique weapons and types of weapons present in the game.

## Changes

* Migrations to create `CanonicalWeapon` model and associations to Enchantments, smithing materials, and tempering materials
* `CanonicalWeapon` model and relevant specs
* JSON data of all canonical weapons
* Rake task to populate/sync canonical weapons
* Update docs on canonical models
* Add possible strength units to the `Enchantment` model to reflect the fact that enchantment strengths may be measured in `"second"`s or in the `"level"` of affected targets
* Add missing materials used for smithing special weapons

### Required Changes

## Considerations

Just like some of our "canonical materials" are actually ingredients, some are now weapons as well. We've opted to include these items, with their normal item codes, as materials in addition to their actual item type for the sake of associations.


lib/tasks/sync_canonical_models.rake
```diff
@@ -123,14 +123,17 @@ namespace :canonical_models do
       Rails.logger.info 'Syncing canonical materials...'
 
       args.with_defaults(preserve_existing_records: false)
+      preserve_existing_records = FALSEY_VALUES.exclude?(args[:preserve_existing_records])
```
danascheider, line 126: There is a [card](https://trello.com/c/K8o44aNI/170-refactor-and-add-tests-for-rake-tasks-syncing-canonical-models) to refactor and add tests for these Rake tasks, so for now I'm letting them be ugly.

PR #94, opened 2022-05-14
Refactor and add tests for Rake tasks syncing canonical models

danascheider:
## Context

[**Refactor and add tests for Rake tasks syncing canonical models**](https://trello.com/c/K8o44aNI/170-refactor-and-add-tests-for-rake-tasks-syncing-canonical-models)

We have some Rake tasks that sync canonical models in the database with the authoritative JSON data kept in this repo. Rake tasks add models that don't exist in the database along with their associations. If the `preserve_existing_records` argument is set to `false` - its default value - then models and associations that aren't in the JSON data will also be removed from the database.

Until this point, these Rake tasks have been untested, and the code in the tasks has gotten progressively more complex as time goes on. It was time to refactor the tasks to use application code that is modular and testable, calling this code in the Rake tasks themselves.

## Changes

* `Canonical::Sync` module providing syncers to sync each type of canonical model
* Tweaks to JSON data to work with new syncers
* RSpec tests for syncers
* Use syncers in Rake tasks that sync models

### Required Changes

## Considerations

It was a little awkward deciding where this `Canonical::Sync` module really fit. I wanted to include it in `lib`, however, I ultimately chose `app/models` because each syncer needs the whole Rails environment loaded, which lib modules don't provide out of the box.

This PR includes some data standardisations: all canonical data now includes an `"attributes"` key, even if there are no other keys (i.e., associations) in the objects. Additionally, all item codes are now uppercase and enchantable item values have been modified on enchantments to correspond to actual types of items (and to add missing types to the lists).


app/models/alchemical_property.rb
```diff
@@ -1,9 +1,15 @@
...
-  has_many :canonical_ingredients, through: :alchemical_properties_canonical_ingredients
+  has_many :canonical_ingredients_alchemical_properties,
+           dependent:  :destroy,
+           class_name: 'Canonical::IngredientsAlchemicalProperty'
+  has_many :canonical_ingredients, through: :canonical_ingredients_alchemical_properties
 
   validates :name, presence: true, uniqueness: { message: 'must be unique' }
   validates :strength_unit, inclusion: { in: %w[point percentage], message: 'must be "point" or "percentage"', allow_blank: true }
+
+  def self.unique_identifier
```
danascheider, line 12: This method is used to determine whether items should be identified by name or item code (or some other attribute) in syncers.

app/models/canonical/sync/association_syncer.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module Canonical
+  module Sync
+    class AssociationSyncer < Syncer
```
danascheider, line 5: The code in this class is ugly and has major metaprogramming energy, however, because of the complexity, generalising and extracting it seemed like the lesser evil compared to having similarly complex but less metaprogrammy code spread out across several classes.

PR #96, opened 2022-05-15
Extract join table for craftable items into polymorphic association

danascheider:
## Context

[**Use polymorphic join table for smithing and tempering materials**](https://trello.com/c/aKETsnue/174-use-polymorphic-join-tables-for-smithing-and-tempering-materials)

In #95, we experimented with using a polymorphic join table to join Enchantments with enchantable items. This made sense because all existing join tables had the same attributes: foreign keys of the relevant models and a `strength` value - the strength of the enchantment on that given model. The experiment was a success, resulting in fewer tables and simplified code.

The same basic pattern applies with both smithing and tempering materials: all join tables include foreign keys and a `quantity` value for the quantity of the material needed. For that reason, we've decided to also use polymorphic join tables for these models.

## Changes

* Migration to create `canonical_craftables_crafting_materials` table
* Remove migrations creating other join tables for smithing materials
* Update model code to account for new join table structure
* Change JSON data to account for new association name `crafting_materials` (instead of `canonical_materials` or `smithing_materials`)
* Update tests, add and remove appropriate specs and factories
* Update Rubocop

### Required Changes

## Considerations

Like in the previous PR, I've decided to modify existing migrations instead of adding a new one. Since these migrations haven't yet run in production, it simplified things a lot to just change it at the point where the tables were originally created, which also enabled us to remove a few other migrations.


app/models/canonical/craftables_crafting_material.rb
```diff
@@ -0,0 +1,17 @@
...
+
+module Canonical
+  class CraftablesCraftingMaterial < ApplicationRecord
+    self.table_name = 'canonical_craftables_crafting_materials'
+
+    belongs_to :craftable, polymorphic: true
+    belongs_to :material, class_name: 'Canonical::Material'
+
+    # rubocop:disable Lint/RedundantCopDisableDirective
+    # rubocop:disable Rails/UniqueValidationWithoutIndex
```
danascheider, line 11: So this is fucking weird and I hate it, but there was a weird disparity between local and CI Rubocop behaviour where, if I didn't disable `Rails/UniqueValidationWithoutIndex`, the cop would fail locally even though the unique index exists (and is created in exactly the same way as the index on  the `canonical_enchantables_enchantments` table, which doesn't produce the same failure). However, in CI, Rubocop determined that this failure would not occur and that the disablement was, therefore, redundant. Thus, I had to also disable that cop, and I had to do it in a separate comment before the one that was deemed redundant.

I tried without success to determine the reason for the disparity between the two models, as well as that for the disparity between local and CI environments. In the pre-commit hook, we do use the command line Rubocop tool rather than the Rails rake task that is used in CI. However, running the Rake task locally results not only in the same failure as the precommit hook finds, but also other failures that don't occur either in CI or in the precommit hook. I'm stumped.

danascheider: I found that, for some reason, although migrations weren't raising any errors, the composite index specified in the migration file was not actually created in the schema. Rolling back the migration and trying again seemed to update the schema so the Rubocop disablement is no longer necessary :tada:

PR #100, opened 2022-05-23
Enable exporting CSVs from JSON

danascheider:
## Context

We've been collecting data in spreadsheets and generating JSON data from the spreadsheets. However, at this point, the JSON data are more complete and accurate than the data in the spreadsheets and should be considered authoritative. Sometimes, nevertheless, it is useful to have data in tabular format for ease of viewing and editing. For these reasons, it made sense to create Rake tasks that would export data from JSON files into CSVs. This way, the data can be re-imported into spreadsheets.

## Changes

* Alphabetise all JSON arrays by name or title if they aren't already
* Add missing "Savior's Hide" item to armour data
* Rake tasks to convert JSON files into CSVs
* Gitignore CSVs in the `/lib/tasks/canonical_models` directory
* Update docs on canonical models

### Required Changes

## Considerations

In some cases, data has to be manipulated somewhat in order to be rendered in tabular format. In particular, nested objects need to be destructured. I've used the `<association>_codes` and `<association>_<attribute>s` headers for this. These columns should consist of a value or comma-separated list of values. Arrays in JSON data, such as `"smithing_perks"` on the `Canonical::Weapon` model, should be likewise joined into comma-separated strings. Note that other separators may be used if the field in question could contain commas. The only current example of this is the `title_variants` field on the `Canonical::Book` model, which is separated using semicolons instead.

I've opted not to include tasks for converting the CSVs back into JSON - mainly because I didn't want to spend the time on un-carded work.

I considered whether I should, again, extract modules to do the heavy lifting on these tasks and test them. Again, I didn't want to spend too much time on uncarded work, so I manually tested the tasks. In the future we may want to extract the code into modules that can be tested.


.gitignore
```diff
@@ -35,3 +35,6 @@ skyrim_inventory_management.csr
 /vendor/bundle/
 
 .pgdata
+
+# Ignore CSV files exported from canonical data (or imported to generate canonical data)
+/lib/tasks/canonical_models/*.csv
```
danascheider, line 40: In general these should not be committed, especially since they can always be generated again with a single Rake task.

lib/tasks/canonical_models/canonical_books.json
```diff
@@ -8319,7 +8319,7 @@
     "attributes": {
       "title": "Note",
       "title_variants": null,
-      "item_code": "0006F63C",
+      "item_code": "0006DFAF",
```
danascheider, line 8322: Changes to this file were presumably the result of sorting the data by title. Notice that these items have the same title and other details, so presumably the sorting algorithm swapped them for some reason, making it seem as though the item codes alone were swapped.

lib/tasks/canonical_models/canonical_jewelry.json
```diff
@@ -3624,6 +3605,25 @@
     ],
     "crafting_materials": []
   },
+  {
+    "attributes": {
```
danascheider, line 3609: Presumably another result of sorting.

lib/tasks/canonical_models/canonical_materials.json
```diff
@@ -172,11 +181,11 @@
   },
   {
     "attributes": {
-      "name": "Small Antlers",
-      "item_code": "0006BC0B",
+      "name": "Troll Skull",
```
danascheider, line 184: I guess this one was added too since it doesn't appear to have been removed from elsewhere in the file during sorting.

lib/tasks/export_csvs.rake
```diff
@@ -0,0 +1,267 @@
...
+  end
+
+  output_value
+end
+
+namespace :csv do
+  namespace :export do
+    desc 'Export a CSV of canonical alchemical properties from JSON data'
+    task alchemical_properties: :environment do
+      json_path = Rails.root.join('lib', 'tasks', 'canonical_models', 'alchemical_properties.json')
```
danascheider, line 21: There's a lot of duplication in this file but I'm just calling it good for now.

PR #102, opened 2022-05-24
Add description to alchemical properties

danascheider:
## Context

[**Add description to canonical alchemical models**](https://trello.com/c/ELOCEVVQ/178-add-description-to-canonical-alchemical-properties)

Alchemical properties have no way to determine what their effects are from the models themselves. It makes sense to add a 'description' field to them so they know their own effects.

## Changes

* Add `description` column to alchemical_properties table
* Add validation for description to `AlchemicalProperty` model
* Add `'level'` as a possible strength unit
* Update spec file to be more consistent with other spec files in addition to testing new functionality

### Required Changes

## Considerations

Descriptions of potions and poisons include information about the strength and duration of their effects (measured by the strength unit of the alchemical property and seconds, respectively). For the canonical model, I've put these variables in angle brackets to enable easy regex replacement. In the models representing potions/poisons, these values will be defined specifically and interpolated. So, "Target is <strength>% weaker to magic for <duration> seconds." becomes "Target is 20% weaker to magic for 30 seconds.", for example, if that is the strength and duration of that particular poison.

### DIFF
--- a/app/models/alchemical_property.rb
+++ b/app/models/alchemical_property.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AlchemicalProperty < ApplicationRecord
+  VALID_STRENGTH_UNITS = %w[point percentage level].freeze
+
   has_many :canonical_ingredients_alchemical_properties,
            dependent:  :destroy,
            class_name: 'Canonical::IngredientsAlchemicalProperty'
@@ -8,7 +10,7 @@ class AlchemicalProperty < ApplicationRecord
 
   validates :name, presence: true, uniqueness: { message: 'must be unique' }
   validates :description, presence: true
-  validates :strength_unit, inclusion: { in: %w[point percentage], message: 'must be "point" or "percentage"', allow_blank: true }
+  validates :strength_unit, inclusion: { in: VALID_STRENGTH_UNITS, message: 'must be "point", "percentage", or the "level" of affected targets', allow_blank: true }
 
   def self.unique_identifier
     :name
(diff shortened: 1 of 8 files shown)


app/models/alchemical_property.rb
```diff
@@ -1,13 +1,16 @@
...
+
   has_many :canonical_ingredients_alchemical_properties,
            dependent:  :destroy,
            class_name: 'Canonical::IngredientsAlchemicalProperty'
   has_many :canonical_ingredients, through: :canonical_ingredients_alchemical_properties
 
   validates :name, presence: true, uniqueness: { message: 'must be unique' }
-  validates :strength_unit, inclusion: { in: %w[point percentage], message: 'must be "point" or "percentage"', allow_blank: true }
+  validates :description, presence: true
+  validates :strength_unit, inclusion: { in: VALID_STRENGTH_UNITS, message: 'must be "point", "percentage", or the "level" of affected targets', allow_blank: true }
```
danascheider, line 13: Turns out some poisons affect characters up to a certain level.

lib/tasks/canonical_models/alchemical_properties.json
```diff
@@ -2,384 +2,439 @@
...
       "name": "Fortify Enchanting",
+      "description": "For <duration> seconds, items are enchanted <strength>% stronger.",
       "strength_unit": "percentage",
       "effects_cumulative": false
     }
   },
   {
     "attributes": {
       "name": "Fortify Health",
-      "strength_unit": "percentage",
```
danascheider, line 103: Oops! Not very many errors in our data but now there's one less.

spec/models/alchemical_property_spec.rb
```diff
@@ -6,32 +6,43 @@
   describe 'validations' do
     describe 'name' do
       it 'must be present' do
-        property = described_class.new
+        model = build(:alchemical_property, name: nil)
```
danascheider, line 9: This is the style of the other spec files, so I've made some changes to this one to bring it in line.

PR #103, opened 2022-05-25
Create Canonical::Potion model

danascheider:
## Context

[**Create Canonical::Potion model**](https://trello.com/c/mbtCSWDA/165-create-canonicalpotion-model)

The last canonical model we need to create as part of the Update Object Modelling epic is `Canonical::Potion`. This model will collate all potions and poisons available in Skyrim.

## Changes

* Migrations to create `canonical_potions` and `canonical_potions_alchemical_properties` tables
* `Canonical::Potion` and `Canonical::PotionsAlchemicalProperties` models
* Specs for the above
* A couple small refactors
* Corrections in JSON data
* JSON data for canonical potions and their properties

### Required Changes

* [ ] Added and updated RSpec tests as appropriate

## Considerations

There are countless possible potions in the game due to player-created potions, which can have a nearly infinite combination of strengths, durations, and effects resulting from player level, perks, and multiple matching properties of ingredients. Consequently, I decided the canonical potion model will only include potions that are purchased or found in the world, not those that can be crafted. Potions that can be crafted will be validated using properties of canonical alchemical ingredients.

### DIFF
--- a/lib/tasks/canonical_models/alchemical_properties.json
+++ b/lib/tasks/canonical_models/alchemical_properties.json
@@ -209,6 +209,14 @@
   },
   {
     "attributes": {
+      "name": "Fortify Speech",
+      "description": "+<strength> Speechcraft for <duration> seconds.",
+      "strength_unit": "point",
+      "effects_cumulative": false
+    }
+  },
+  {
+    "attributes": {
       "name": "Fortify Stamina",
       "description": "Stamina is increased by <strength> points for <duration> seconds.",
       "strength_unit": "point",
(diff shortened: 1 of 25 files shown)


lib/tasks/canonical_models/alchemical_properties.json
```diff
@@ -207,6 +207,14 @@
       "effects_cumulative": false
     }
   },
+  {
+    "attributes": {
+      "name": "Fortify Speech",
```
danascheider, line 212: This property doesn't exist on any ingredients (except through [Creations](https://en.uesp.net/wiki/Skyrim:Creation_Club)) but there are potions in-game with these effects.

lib/tasks/export_csvs.rake
```diff
@@ -11,7 +11,7 @@ def concatenate_values(array, *keys_to_concatenate)
     output_value[index] = values.empty? ? nil : values.join(',')
   end
 
-  output_value
+  output_value.map(&:presence)
```
danascheider, line 14: This was a bug - empty values should be `nil`, not `''`.

PR #104, opened 2022-05-25
Refactor/reformat syncer specs a bit

danascheider:
## Context

In the process of QAing the Update Object Modelling initiative, I realised the code for the syncer specs was a bit ugly and not all the tests did what they should.

## Changes

* Light refactoring of syncer specs
* Reformatting of specs to remove excessively long lines or repeated creation of associations in `before` blocks

### Required Changes

## Considerations

There's a lot of more or less repeated code in the syncer specs. I tried extracting it to a shared example group, but determined there were just slightly too many things specific to each spec to make that practical (model name, names of associations, fixture data, etc. in addition to memoised helpers already defined).


spec/models/canonical/sync/alchemical_properties_spec.rb
```diff
@@ -30,9 +30,12 @@
       end
 
       context 'when there are no existing records in the database' do
-        it 'populates the models from the JSON file' do
+        it 'populates the models from the JSON file', :aggregate_failures do
           perform
-          expect(AlchemicalProperty.count).to eq 4
+          expect(AlchemicalProperty.find_by(name: 'Cure Disease')).to be_present
```
danascheider, line 35: An attentive reader will notice that, in some of the spec files, it's just `expect { perform }.to change(Model, :count).from(0).to(4)` and in some, such as this one, the specific models are actually expected. The reason for this is that, in the case of models that have associations, the existence of the specific models is implicitly tested in the tests for their associations. When models don't have associations, there aren't other specs where the model's existence is tested so we need to test for them explicitly.

spec/support/fixtures/canonical/sync/books.json
```diff
@@ -1,4 +1,28 @@
 [
+  {
+    "attributes": {
+      "title": "Cure Disease Potion Recipe",
```
danascheider, line 4: I moved this object to be the first one in the JSON array since it is used as the `item_in_json` for the specs, and for all the other specs the `item_in_json` is the first one in the fixture file.

sim-api.danascheider.com.crt
```diff
@@ -1,31 +0,0 @@
------BEGIN CERTIFICATE-----
```
danascheider, line 1: Should never have been committed to Git but OK

PR #108, opened 2022-05-27
Add fields to canonical armors

danascheider:
## Context

[**Add boolean columns to canonical models that don't have them**](https://trello.com/c/rNjxj88l/179-add-boolean-columns-to-canonical-models-that-dont-have-them)

Canonical models representing in-game items should all have four attributes:

* `purchasable`: Whether the item can be purchased from merchants or other NPCs
* `unique_item`: Whether the item is unique in the game (items that respawn can still be unique as long as they aren't found in any other locations in the game)
* `rare_item`: Whether an item is rare*
* `quest_item`: Whether an item is (a) required to complete a quest or (b) only obtainable through completing a quest

Additionally, since `Canonical::Weapon` models have a `smithing_perks` field (indicating which smithing perks the items require or benefit from), I figured `Canonical::Armor` should too.

*Rare items include non-purchasable items that are found in 10 or fewer locations in the game, purchasable items that are found in two or fewer other locations in the game, or items that are theoretically purchasable but are not routinely a part of any particular merchant's inventory.

## Changes

* Migration to add `smithing_perks` array column to `canonical_armors` table
* Update schema to reflect changes made in previous migrations
* Add all missing values to each item in the JSON data for canonical armors
* Update specs, factories, and fixtures
* Add validations and specs

### Required Changes

## Considerations

Although the new fields should all be `NOT NULL` (as should all boolean fields, in fact), creating a constraint now would make it impossible to migrate the database in production since existing models won't have these values defined. However, since we want all new models to include them, I have added validations to the model class.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_05_26_074029) do
+ActiveRecord::Schema.define(version: 2022_05_26_213259) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -38,6 +38,7 @@ ActiveRecord::Schema.define(version: 2022_05_26_074029) do
     t.boolean "enchantable", default: true
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.string "smithing_perks", default: [], array: true
     t.index ["item_code"], name: "index_canonical_armors_on_item_code", unique: true
   end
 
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -33,11 +33,15 @@
     t.string "magical_effects"
     t.decimal "unit_weight", precision: 5, scale: 2, null: false
     t.boolean "dragon_priest_mask", default: false
-    t.boolean "quest_item", default: false
-    t.boolean "unique_item", default: false
     t.boolean "enchantable", default: true
+    t.boolean "leveled", default: false
```
danascheider, line 37: I have no idea why these fields weren't in the schema before because they were definitely in the  migrations - and were additionally somehow present in the production database (but not my dev environment, even after I dropped and re-setup the database).

PR #109, opened 2022-05-28
Add missing booleans to canonical clothing items data

danascheider:
## Context

[**Add missing booleans to canonical clothing items data**](https://trello.com/c/IXZefWFa/181-add-missing-booleans-to-canonical-clothing-items-data)

We've added the following four fields to any canonical models where they make sense:

* `purchasable`: whether or not the item can be purchased from a merchant or other NPC
* `unique_item`: whether the item is unique in the game (includes items that are available in only one location but respawn)
* `rare_item`: whether the item is rare* in the game
* `quest_item`: whether the item is (a) required to complete a quest or (b) only obtainable by completing a quest

Since we decided late in the game (no pun intended) that we wanted these fields to be common to all canonical models, adding them as required fields like we wanted would entail adding the fields to the database, updating all the existing JSON data, loading the JSON data, and then adding the `NOT NULL` constraints. Since we didn't want the considerable work of updating the JSON data to block release of the [Update Object Modelling epic](https://trello.com/c/1G99z7sP/68-update-object-modelling) (feature branch #77) or MVP, we decided to release the epic without the `NOT NULL` constraints and gradually update JSON data as there was time. Luckily, updating the JSON data is not taking as long as expected.

In order to keep migrations to a minimum, we'll be adding the `NOT NULL` constraints in a separate PR, after all JSON data have been updated. However, it makes sense to add validations to models so that any new data added to the database, as well as any updated records, must include the fields.

*Rare items are defined, generally, as those that are not purchasable and available in fewer than 10 locations in the game or purchasable and available two or fewer other places in the game. There are a few exceptions to this, such as items whose few locations are so obvious players are all but guaranteed to find them. In particular, if any of an item's locations includes an ownable property (Breezehome, Lakeview Manor, etc.), it will not be considered rare.

## Changes

* Update JSON data for canonical clothing items to include missing fields
* Add validations to models to ensure fields aren't blank
* Update tests

### Required Changes

### DIFF
--- a/spec/models/canonical/clothing_item_spec.rb
+++ b/spec/models/canonical/clothing_item_spec.rb
@@ -5,73 +5,133 @@ require 'rails_helper'
 RSpec.describe Canonical::ClothingItem, type: :model do
   describe 'validations' do
     it 'is valid with valid attributes' do
-      item = described_class.new(name: 'Clothes', item_code: 'foo', unit_weight: 1, body_slot: 'body')
+      model = described_class.new(
+                name:        'Clothes',
+                item_code:   'foo',
+                unit_weight: 1,
+                body_slot:   'body',
+                purchasable: true,
+                unique_item: false,
+                rare_item:   false,
+              )
 
-      expect(item).to be_valid
+      expect(model).to be_valid
     end
 
     describe 'name' do
-      it 'is invalid without a name' do
-        item = described_class.new(item_code: 'xxx', body_slot: 'hands', unit_weight: 4.2)
+      it "can't be blank" do
+        model = build(:canonical_clothing_item, name: nil)
 
-        item.validate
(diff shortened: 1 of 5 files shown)


spec/models/canonical/clothing_item_spec.rb
```diff
@@ -5,73 +5,133 @@
...
+                rare_item:   false,
+              )
+
+      expect(model).to be_valid
     end
 
     describe 'name' do
-      it 'is invalid without a name' do
-        item = described_class.new(item_code: 'xxx', body_slot: 'hands', unit_weight: 4.2)
+      it "can't be blank" do
```
danascheider, line 22: Updated the existing specs to be more similar to other spec files.

PR #112, opened 2022-05-29
Add new fields and validations to Canonical::Ingredient model

danascheider:
## Context

[**Add missing booleans to canonical ingredients data**](https://trello.com/c/c0tsey38/182-add-missing-booleans-to-canonical-ingredients-data)

We want all canonical models for which it makes sense to have a common API with fields for `purchasable`, `unique_item`, `rare_item`, and `quest_item`. These columns were previously added to the `canonical_ingredients` table (without constraints), but the data have not been updated yet.

In the course of updating the data, it was determined that there were other useful fields canonical ingredients should include:

* `ingredient_type`: indicates merchant availability
* `purchase_requires_perk`: indicates whether the Merchant perk is required for an item to be available for purchase

## Changes

* Migration to add `ingredient_type` and `purchase_requires_perk` fields to `canonical_ingredients` table
* Updated test fixtures and JSON data for syncers
* Validations on the `Canonical::Ingredient` model testing for inclusion of required fields

### Required Changes

## Considerations

The `purchasable` column, as discussed in the added docs, is probably redundant on canonical ingredients, as purchasability can be determined by the value, or lack thereof, of the `ingredient_type` column. It would've probably been just as good to drop the `purchasable` column and simply define a `#purchasable` method on the model to maintain the common API. However, this is more easily done than undone, and we'll know more about the requirements of this table once we've started the [ingredients epic](https://trello.com/c/WE1ztpCb/154-ingredient-features), so I've decided to leave it be for now.

Also discussed in the new docs are the complications involved in identifying an ingredient as a `rare_item` or not. Since ingredients are consumable, an ingredient available in less than 10 locations in the game would be rare indeed. Additionally, ingredients are never guaranteed to be available from a particular merchant at a particular time (or, indeed, ever). For this reason, I've taken a slightly different approach to categorising ingredients in this way, which takes into account several factors, including the subjective experience of how often I've come across the ingredient in the game when I've played it.

### DIFF
--- a/app/models/canonical/ingredient.rb
+++ b/app/models/canonical/ingredient.rb
@@ -4,6 +4,11 @@ module Canonical
   class Ingredient < ApplicationRecord
     self.table_name = 'canonical_ingredients'
 
+    VALID_TYPES                = %w[common uncommon rare Solstheim].freeze
+    TYPE_VALIDATION_MESSAGE    = 'must be "common", "uncommon", "rare", or "Solstheim"'
+    BOOLEAN_VALUES             = [true, false].freeze
+    BOOLEAN_VALIDATION_MESSAGE = 'must be true or false'
+
     has_many :canonical_ingredients_alchemical_properties,
              dependent:  :destroy,
              class_name: 'Canonical::IngredientsAlchemicalProperty'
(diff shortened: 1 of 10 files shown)


app/models/canonical/ingredient.rb
```diff
@@ -17,10 +22,39 @@ class Ingredient < ApplicationRecord
...
     validates :item_code, presence: true, uniqueness: { message: 'must be unique' }
+    validates :ingredient_type, inclusion: { in: VALID_TYPES, message: TYPE_VALIDATION_MESSAGE, allow_blank: true }
     validates :unit_weight, presence: true, numericality: { greater_than_or_equal_to: 0 }
+    validates :purchase_requires_perk,
+              inclusion: {
+                           in:      BOOLEAN_VALUES,
+                           message: "#{BOOLEAN_VALIDATION_MESSAGE} if purchasable is true",
+                         },
+              if:        -> { purchasable == true }
+    validates :unique_item, inclusion: { in: BOOLEAN_VALUES, message: BOOLEAN_VALIDATION_MESSAGE }
```
danascheider, line 33: The fact `unique_item` has a default value prevents this from ever really coming up. If the value is `nil` it will be automatically set to `false` prior to validation, and truthy values will be set to `true` prior to validation whether there is a default value or not.

PR #115, opened 2022-05-31
Add missing booleans to canonical weapons data

danascheider:
## Context

[**Add missing booleans to canonical weapons data**](https://trello.com/c/ysHroToa/185-add-missing-booleans-to-canonical-weapons-data)

We've standardised the API for canonical models somewhat, so each model for which it makes sense will include the following four fields:

* `purchasable`: whether the item is available for purchase from merchants or other NPCs
* `unique_item`: whether the item is unique in the game (includes items that respawn, as long as they are available from only one location in the game and that location is not a vendor)
* `rare_item`: whether the item is rare (as defined below); all unique items are also rare
* `quest_item`: whether the item is (a) required to complete a quest or (b) only obtainable through completing a quest

Rare items are defined, generally, as those that (a) are not purchasable and are found in fewer than 10 locations in the game or (b) are purchasable and are found in fewer than 3 locations in the game other than a vendor. ((b) does not apply if the item is consistently or frequently available from a variety of vendors.) For purposes of classifying rare items, random loot and random drops each count as half of a fixed location, since they are a possible place to find the item but are not guaranteed. The exception is if they are always or frequently dropped by a common type of enemy. Note that, in the weapons category, arrows are treated somewhat differently, since they are consumable. They've been classified as rare based on the subjective experience of finding them in the game.

The `canonical_weapons` table already has all four of these fields, however, they are not all present in the JSON data for weapons. Since the goal is to make these fields required (`NOT NULL`), we need to add the fields to the JSON data and sync the prod database before we can add `NOT NULL` constraints.

## Changes

* Add values for `purchasable` and `rare_item` to canonical weapons data
* Add validations to the `Canonical::Weapon` model and tests for those validations
* Update factory and fixture data
* Add `Power` as a prerequisite for the canonical weapon syncer (this is an oversight from a previous PR)

### Required Changes


app/models/canonical/weapon.rb
```diff
@@ -75,9 +77,16 @@ class Weapon < ApplicationRecord
                          }
     validates :base_damage, presence: true, numericality: { greater_than_or_equal_to: 0, only_integer: true }
     validates :unit_weight, presence: true, numericality: { greater_than_or_equal_to: 0 }
+    validates :purchasable, inclusion: { in: BOOLEAN_VALUES, message: BOOLEAN_VALIDATION_MESSAGE }
+    validates :unique_item, inclusion: { in: BOOLEAN_VALUES, message: BOOLEAN_VALIDATION_MESSAGE }
+    validates :rare_item, inclusion: { in: BOOLEAN_VALUES, message: BOOLEAN_VALIDATION_MESSAGE }
+    validates :quest_item, inclusion: { in: BOOLEAN_VALUES, message: BOOLEAN_VALIDATION_MESSAGE }
+    validates :leveled, inclusion: { in: BOOLEAN_VALUES, message: BOOLEAN_VALIDATION_MESSAGE }
+    validates :enchantable, inclusion: { in: BOOLEAN_VALUES, message: BOOLEAN_VALIDATION_MESSAGE }
```
danascheider, line 85: All these fields should be validated, not just the new ones.

PR #119, opened 2022-06-08
Update hash syntax to use value omission

danascheider:
## Context

[**Update to the latest Ruby**](https://trello.com/c/RjRNHErC/192-update-to-the-latest-ruby)

We've updated Ruby to 3.1.2 in #118 but still need to make changes to make use of its features and write code in a style more suited to Ruby 3.1. This includes, at the most basic level, fixing Rubocop failures that have cropped up, but also making use of expanded functionality. This PR includes only updating hash syntax to use value omission. For example:

```ruby
# Ruby 3.0
list_id = 4
hash = { list_id: list_id }

# Ruby 3.1
hash = { list_id: }
```

## Changes

* Update hash syntax to use value omission where possible (e.g., `{ list_id: }` instead of `{ list_id: list_id }`)

### Required Changes

## Considerations

I've created this PR to only update hash syntax because there were so many files that needed to be changed for that change alone. All corrections were done automatically using Rubocop:

```sh
bundle exec rubocop -A --only Style/HashSyntax
```

### Notes

After consideration, I've decided this will be the only PR to make use of Ruby 3.1 features. The reason is that other features are harder to implement using a simple process of grepping through the code base for key words and making changes where I find them, so instead, I'll take the approach of updating files, if appropriate, as they are touched in the course of other work.

### DIFF
--- a/spec/models/canonical/ingredients_alchemical_property_spec.rb
+++ b/spec/models/canonical/ingredients_alchemical_property_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe Canonical::IngredientsAlchemicalProperty, type: :model do
           )
         end
 
-        new_association = build(:canonical_ingredients_alchemical_property, ingredient: ingredient)
+        new_association = build(:canonical_ingredients_alchemical_property, ingredient:)
 
         new_association.validate
         expect(new_association.errors[:ingredient]).to include 'already has 4 alchemical properties'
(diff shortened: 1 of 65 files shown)


spec/models/canonical/ingredients_alchemical_property_spec.rb
```diff
@@ -30,15 +30,15 @@
           create(
             :canonical_ingredients_alchemical_property,
             priority:   1,
```
danascheider, line 32: I don't like that this maintains the extra 2 spaces to theoretically align values even though the value is omitted on the next line. I've found lots of issues with this Rubocop cop in general and am thinking of changing the required style to _not_ align values in these situations, but that's a change for another PR.

PR #144, opened 2023-03-10
Reimplement Sign in with Google via Firebase

danascheider:
## Context

* [**Implement Sign in with Google via Firebase**](https://trello.com/c/a003Z9Gj/229-implement-sign-in-with-google-via-firebase)
* [**Rewrite Games page**](https://trello.com/c/P1WhcSOq/231-rewrite-games-page)

We have reached the point in the front end rewrite where we need the backend to correctly validate access tokens returned to the front end from Google. This PR adds back token validation, re-enables multiple users, and ensures that multi-user and authentication functionality are fully tested.

## Changes

* Rename columns in the Users table to match values from Google
* Update code and factories to reflect new column names
* Remove email uniqueness validation from users table since that's not how we're uniquely identifying users anymore
* Add `dotenv` gem to manage environment variables in dev environments
* Use Faraday to validate ID tokens on authenticated requests
* Ensure that models belonging to users are no longer automatically configured to belong to only the first user
* Add WebMock to enable network-level stubbing of authentication
* Update tests to account for multi-user factors and authentication

### Required Changes

## Considerations

Because of severely lacking Google documentation, it's not clear what an unsuccessful response to token validation would look like. For that reason, we weren't able to test the `AuthorizationService` as thoroughly as would be desirable. This implementation deems a token adequately validated if the response status from the validation endpoint is in the 200 range and the shape of the response body matches expectations (i.e., JSON object that includes a `"users"` array with exactly one member).


app/controller_services/application_controller/authorization_service.rb
```diff
@@ -1,26 +1,51 @@
...
 
-      controller.current_user = User.first
-      nil
+        raise AmbiguousUserError.new('Token validation response did not include a user') if users.blank?
+        raise AmbiguousUserError.new('Token validation response included multiple users') if users.length > 1
+
+        controller.current_user = User.create_or_update_for_google(users.first)
+        nil
+      else
+        Rails.logger.debug token_response.body
```
danascheider, line 30: This is kind of temporary; we've never seen what such an error response might look like and would like to "capture" one for testing purposes.

spec/support/auth_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+        status: 200,
+        body:   File.read(Rails.root.join('spec', 'support', 'fixtures', 'auth', 'success.json')),
+      )
+  end
+
+  def stub_unsuccessful_login
+    stub_request(:post, auth_uri)
+      .to_return(
+        status: 200,
+        body:   File.read(Rails.root.join('spec', 'support', 'fixtures', 'auth', 'empty_users_array.json')),
```
danascheider, line 20: Picked an invalid fixture arbitrarily.

PR #150, opened 2023-03-15
Return all remaining shopping lists from DELETE /shopping_lists/:id endpoint

danascheider:
## Context

[**Return all shopping lists from DELETE /shopping_lists/:id endpoint**](https://trello.com/c/nb0KVUOv/258-return-all-shopping-lists-from-delete-shoppinglists-id-endpoint)

We're trying to simplify API response bodies so the front end doesn't require as much conditional logic to handle them. One part of this is changing the shopping list deletion endpoint so that, instead of a possible 204 response, it always returns a 200 response with an array of any remaining shopping lists belonging to the same game.

## Changes

* Return a uniform response including all remaining shopping lists belonging to the same game when a shopping list is destroyed
* Update tests
* Update API docs

### Required Changes

### DIFF
--- a/spec/controller_services/shopping_lists_controller/destroy_service_spec.rb
+++ b/spec/controller_services/shopping_lists_controller/destroy_service_spec.rb
@@ -18,10 +18,14 @@ RSpec.describe ShoppingListsController::DestroyService do
       let(:game) { create(:game, user:) }
 
       context 'when the game has additional regular lists' do
-        let!(:third_list) { create(:shopping_list, game:, aggregate_list:) }
+        let!(:third_list) { create(:shopping_list_with_list_items, game:, aggregate_list:) }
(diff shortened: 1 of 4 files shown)


spec/controller_services/shopping_lists_controller/destroy_service_spec.rb
```diff
@@ -18,12 +18,16 @@
       let(:game) { create(:game, user:) }
 
       context 'when the game has additional regular lists' do
-        let!(:third_list) { create(:shopping_list, game:, aggregate_list:) }
+        let!(:third_list) { create(:shopping_list_with_list_items, game:, aggregate_list:) }
```
danascheider, line 21: Might as well throw some additional list items into the mix.

PR #155, opened 2023-03-21
Return IDs of deleted lists from DELETE /shopping_lists/:id endpoint

danascheider:
## Context

[**Return IDs of deleted lists from DELETE /shopping_lists/:id endpoint**](https://trello.com/c/611KOMJq/264-return-ids-of-deleted-lists-from-delete-shoppinglists-id-endpoint)

We've realised some issues with the implementation of our endpoints in that the new response bodies added in #150 are causing shopping lists to be rearranged too much on the page due to the `index_order` scope on shopping lists and their items. We've decided to change the response bodies so it's easier for the front end to ensure that the lists stay in their expected position until page reload.

## Changes

* Change response body shape of `DELETE /shopping_lists/:id` endpoint
* Update RSpec tests
* Update docs

### Required Changes

## Considerations

This made the code in the `ShoppingListsController::DestroyService` a little uglier but I didn't really know a better way to implement it.

### DIFF
--- a/app/controller_services/shopping_lists_controller/destroy_service.rb
+++ b/app/controller_services/shopping_lists_controller/destroy_service.rb
@@ -17,9 +17,13 @@ class ShoppingListsController < ApplicationController
     def perform
       return Service::MethodNotAllowedResult.new(errors: [AGGREGATE_LIST_ERROR]) if shopping_list.aggregate == true
 
-      game = shopping_list.game
+      ids = game.shopping_lists.count == 2 ? [aggregate_list.id, shopping_list.id] : [shopping_list.id]
+
       destroy_and_update_aggregate_list_items!
-      Service::OKResult.new(resource: game.shopping_lists.index_order)
+
+      resource = aggregate_list&.persisted? ? { deleted: ids, aggregate: aggregate_list } : { deleted: ids }
+
+      Service::OKResult.new(resource:)
     rescue ActiveRecord::RecordNotFound
       Service::NotFoundResult.new
     rescue StandardError => e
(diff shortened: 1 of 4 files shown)


app/controller_services/shopping_lists_controller/destroy_service.rb
```diff
@@ -35,6 +39,14 @@ def shopping_list
       @shopping_list ||= user.shopping_lists.find(list_id)
     end
 
+    def aggregate_list
+      game.aggregate_shopping_list
```
danascheider, line 43: It's not an accident that I didn't include `@aggregate_list ||=` here - memoising the value actually would precent ActiveRecord from detecting that the list wasn't persisted on line 24.

app/controller_services/shopping_lists_controller/destroy_service.rb
```diff
@@ -35,6 +39,14 @@ def shopping_list
...
+    end
+
+    def game
+      @game ||= shopping_list.game
```
danascheider, line 47: It bothers me a little bit defining this value in this way - it's a bit brittle because if `shopping_list` were to not be called until after the shopping list was deleted, this would raise an error. However, if that happened, the `#shopping_list` method would also raise `ActiveRecord::RecordNotFound`, so it wouldn't even get to the point of this being called.

spec/controller_services/shopping_lists_controller/destroy_service_spec.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
-require 'service/no_content_result'
```
danascheider, line 4: This was a holdover from when `204 No Content` was an actual possible response from this endpoint.

spec/requests/shopping_lists_spec.rb
```diff
@@ -772,8 +772,17 @@
 
       context 'when the shopping list exists' do
         let!(:shopping_list) { create(:shopping_list, game:) }
+        let!(:shopping_list_id) { shopping_list.id }
 
         context "when this is the game's last regular shopping list" do
+          let!(:aggregate_list_id) { game.aggregate_shopping_list.id }
```
danascheider, line 778: There would've been a few ways to skin this cat, including `shopping_list.aggregate_list_id`. I didn't see too many advantages of one approach over another.

PR #156, opened 2023-03-29
Return only changed shopping lists from list item creation endpoint

danascheider:
## Context

* [**Rewrite Shopping Lists page**](https://trello.com/c/oBVGCwU1/232-rewrite-shopping-lists-page)
* [**Determine best approach to updating aggregatable models**](https://trello.com/c/GO0zCmKC/262-determine-best-approach-to-updating-aggregatable-models)

In #151, we updated the shopping list item creation endpoint to return all of a game's shopping lists and their list items. The goal of this was to simplify front-end response handling, however, there was a flaw in this approach: simply returning all the shopping lists and setting them as the `shoppingLists` array on the front end would cause the page to be reorganised frequently, with the most-recently modified lists always being first.

For the shopping list item creation endpoint, we opted to take the middle way: return shopping lists from the endpoint and not just list items like before, but only return the changed shopping lists and not every one for that game. The front end will then be responsible for determining the appropriate order for the returned lists in the display.

## Changes

* Return all changed shopping lists for the relevant game when a shopping list item is created
* Update RSpec tests
* Update docs

### Required Changes

### Drawbacks to This Approach

Although this approach will mitigate the problem of shopping lists being moved around on the page, it will still result in the list items for each list being reorganised. This could be jarring for users if they have multiple lists expanded and can see their list items move around when they submit the form. However, we are going to experiment with this and see how big a deal it really turns out to be before reverting to the original response body that included the list items only. The reason for this is that it is much less computationally expensive on the front end to replace a shopping list in the array than to find the shopping list the list item belongs to and change just one list item.

### Response Status

Currently, the endpoint returns a 201 status if the list item was created, and 200 if it is combined with another item of the same description on the same list. I debated whether to change this because the front end only really cares if the response was successful, since it is just replacing lists with those from the response regardless. However, I couldn't decide which status to use for the single status and thought the distinction might eventually be important, so I left it as-is. If it becomes too annoying or isn't useful in the future we can change it.

### DIFF
--- a/app/controller_services/shopping_list_items_controller/create_service.rb
+++ b/app/controller_services/shopping_list_items_controller/create_service.rb
@@ -24,16 +24,18 @@ class ShoppingListItemsController < ApplicationController
       item = ShoppingListItem.combine_or_new(params.merge(list_id:))
 
       ActiveRecord::Base.transaction do
+        lists_changed = lists_to_be_changed
+
         item.save!
 
         if preexisting_item.blank?
           aggregate_list.add_item_from_child_list(item)
 
-          Service::CreatedResult.new(resource: shopping_list.game.shopping_lists.index_order)
+          Service::CreatedResult.new(resource: lists_changed)
         else
           aggregate_list.update_item_from_child_list(params[:description], params[:quantity], params[:unit_weight], nil, params[:notes])
 
-          Service::OKResult.new(resource: shopping_list.game.shopping_lists.index_order)
+          Service::OKResult.new(resource: lists_changed)
         end
       end
     rescue ActiveRecord::RecordInvalid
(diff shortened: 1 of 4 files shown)


app/controller_services/shopping_list_items_controller/create_service.rb
```diff
@@ -24,16 +24,18 @@ def perform
       item = ShoppingListItem.combine_or_new(params.merge(list_id:))
 
       ActiveRecord::Base.transaction do
+        lists_changed = lists_to_be_changed
```
danascheider, line 27: Unfortunately, this method needs to be called at this point instead of just adding it on lines 34 and 38 like I had hoped. The reason is because, by line 34/38, if the unit weight is being changed, it has already been updated on the aggregate list item and we can no longer tell if it was changed.

app/controller_services/shopping_list_items_controller/create_service.rb
```diff
@@ -57,8 +59,28 @@ def aggregate_list
...
     def all_matching_list_items
-      aggregate_list.game.shopping_list_items.where('description ILIKE ?', params[:description])
+      @all_matching_list_items ||= game.shopping_list_items.where(
+        'description ILIKE ?', params[:description],
+      )
+    end
+
+    def lists_to_be_changed
+      list_ids = if params[:unit_weight] && params[:unit_weight] != aggregate_list_item&.unit_weight
+                   all_matching_list_items.pluck(:list_id).push(shopping_list.id)
```
danascheider, line 78: The list IDs will already include `shopping_list.id` if there is a matching item already on that shopping list. However, there's no real disadvantage to having it on the list twice and it would be more code to check or call `uniq` than to just leave it.

spec/requests/shopping_list_items_spec.rb
```diff
@@ -45,17 +45,21 @@
               expect(response.status).to eq 201
             end
 
-            it 'returns all shopping lists for the same game' do
+            it 'returns all changed shopping lists for the same game' do
               create_item
-              expect(response.body).to eq(game.shopping_lists.reload.index_order.to_json)
+              expect(response.body).to eq(game.shopping_lists.to_json)
```
danascheider, line 50: In this case it's the same - these are all the shopping lists. The `reload` and `index_order` are redundant.

spec/requests/shopping_list_items_spec.rb
```diff
@@ -374,7 +392,7 @@
 
           it "returns all the game's shopping lists" do
             update_item
-            expect(response.body).to eq(game.shopping_lists.reload.index_order.to_json)
```
danascheider, line 377: Realised this was all redundant.

PR #157, opened 2023-03-30
Fix bug where aggregate list not returned from item creation endpoint

danascheider:
## Context

* [**Rewrite Shopping Lists page**](https://trello.com/c/oBVGCwU1/232-rewrite-shopping-lists-page)
* [**Enable shopping list item creation from shopping lists page**](https://trello.com/c/o4ghq9eI/266-enable-shopping-list-item-creation-from-shopping-lists-page)

This PR should've been part of #156, where we missed a critical edge case. We want any changed shopping lists to be returned from the shopping list item creation endpoint. However, we also made the error of returning only the shopping list to which an item was directly added if:

1. The request set a unit weight for the item
2. No matching item existed on the aggregate list

## Changes

* Fix bug where only shopping list where item was directly created is returned if request sets unit weight and no aggregate list item exists
* Add tests covering this regression

### Required Changes

## Considerations

Docs didn't need to be updated because they already stated the desired behaviour - we didn't know the behaviour _wasn't_ the desired behaviour until after they were written.

### DIFF
--- a/app/controller_services/shopping_list_items_controller/create_service.rb
+++ b/app/controller_services/shopping_list_items_controller/create_service.rb
@@ -74,7 +74,7 @@ class ShoppingListItemsController < ApplicationController
     end
 
     def lists_to_be_changed
-      list_ids = if params[:unit_weight] && params[:unit_weight] != aggregate_list_item&.unit_weight
+      list_ids = if all_matching_list_items.count > 0 && params[:unit_weight] && params[:unit_weight] != aggregate_list_item&.unit_weight
                    all_matching_list_items.pluck(:list_id).push(shopping_list.id)
                  else
                    [aggregate_list.id, shopping_list.id]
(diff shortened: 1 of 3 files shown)


app/controller_services/shopping_list_items_controller/create_service.rb
```diff
@@ -74,7 +74,7 @@ def all_matching_list_items
     end
 
     def lists_to_be_changed
-      list_ids = if params[:unit_weight] && params[:unit_weight] != aggregate_list_item&.unit_weight
+      list_ids = if all_matching_list_items.count > 0 && params[:unit_weight] && params[:unit_weight] != aggregate_list_item&.unit_weight
```
danascheider, line 77: I tried a couple approaches to making this line shorter but one caused a syntax error and the other caused Rubocop to do something really weird with the spacing, so I am reluctantly leaving this line very long.

PR #164, opened 2023-04-10
Remove notes values from aggregate shopping list/inventory items

danascheider:
## Context

[**Investigate bug with combining notes values on aggregate lists**](https://trello.com/c/KlCqMnkY/274-investigate-bug-with-combining-notes-values-on-aggregate-lists)

We found some unexpected aggregation behaviour when adding, updating, and deleting shopping list items from the shopping lists page. Notes tended to disappear at seemingly random times from the aggregate list items, instead of being appropriately aggregated. The original card was intended to identify and fix the bugs. However, I ultimately decided to remove `notes` from aggregate list items entirely.

## Changes

* Update `Aggregatable` methods `#update_item_from_child_list`, `#add_item_from_child_list`, and `#remove_item_from_child_list` to ignore `notes` on aggregate items
* Update tests and docs
* Add validations to ensure aggregate list items cannot be created or updated with notes
* One-off Rake task to remove existing `notes` values from aggregate list items

### Required Changes

## Considerations

After some consideration, I realised the benefit of aggregating notes on the aggregate lists is minimal and could even confuse users. These notes are specific to child lists anyway. If we were going to do something similar, it'd be more helpful to create an association between aggregate list items and corresponding child list items, to help users identify where to look for the item.


config/initializers/application_controller_renderer.rb
danascheider, line 1: This file was empty so I just deleted it.

config/initializers/cors.rb
```diff
@@ -9,7 +9,7 @@
 
 Rails.application.config.middleware.insert_before 0, Rack::Cors do
   allow do
-    origins ENV['CLIENT_ORIGIN'] || 'http://localhost:5173'
+    origins ENV['CLIENT_ORIGIN'].presence || 'http://localhost:5173'
```
danascheider, line 12: This ensures that if the env variable is set to `''` the client origin is still a valid URI.

PR #177, opened 2023-05-25
Create non-canonical Ingredient model

danascheider:
## Context

[**Create non-canonical Ingredient model**](https://trello.com/c/LhzhuYpF/306-create-non-canonical-ingredient-model)

We are creating non-canonical models to represent items players can discover in-game. The next model we need to create is `Ingredient`. This model will play a critical role in the upcoming ingredients epic.

## Changes

* Create non-canonical `Ingredient` model
* Create non-canonical `IngredientsAlchemicalProperty` join model
* Add factories for new models
* Add tests for new models
* Update docs

### Required Changes

## Considerations

I decided not to create a direct association between the `Canonical::IngredientAlchemicalProperty` and `IngredientAlchemicalProperty` models. This will slow down validations slightly but I don't currently see a compelling reason to add such an association and it feels like it would complicate things.

### DIFF
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+class Ingredient < ApplicationRecord
+  DOES_NOT_MATCH = "doesn't match an ingredient that exists in Skyrim"
+
+  belongs_to :game
+  belongs_to :canonical_ingredient,
+             class_name: 'Canonical::Ingredient',
+             optional: true,
+             inverse_of: :ingredients
+
+  validates :name, presence: true
+  validate :ensure_match_exists
+
+  before_validation :set_canonical_ingredient
+
+  private
+
+  def set_canonical_ingredient
+    matching = Canonical::Ingredient.where('name ILIKE ?', name)
+
+    self.canonical_ingredient = matching.first if matching.count == 1
+  end
+
+  def ensure_match_exists
+    return if canonical_ingredient.present?
+    return if Canonical::Ingredient.where('name ILIKE ?', name).present?
+
+    errors.add(:base, DOES_NOT_MATCH)
+  end
+end
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -19,12 +19,31 @@ class Ingredient < ApplicationRecord
 
   before_validation :set_canonical_ingredient
 
+  def canonical_ingredients
+    return Canonical::Ingredient.where(id: canonical_ingredient.id) if canonical_ingredient.present?
+
+    matching = Canonical::Ingredient.where('name ILIKE ?', name)
+
+    return matching unless alchemical_properties.any?
+
+    ingredients_alchemical_properties.each do |join_model|
+      matching = matching.joins(:canonical_ingredients_alchemical_properties).where(
+        'canonical_ingredients_alchemical_properties.alchemical_property_id = :property_id AND canonical_ingredients_alchemical_properties.priority = :priority',
+        property_id: join_model.alchemical_property_id,
+        priority: join_model.priority,
+      )
+    end
+
+    matching
+  end
+
   private
 
   def set_canonical_ingredient
-    matching = Canonical::Ingredient.where('name ILIKE ?', name)
+    return if canonical_ingredient.present?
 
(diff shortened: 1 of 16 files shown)


app/models/ingredient.rb
```diff
@@ -0,0 +1,55 @@
...
+           -> { select 'alchemical_properties.*, ingredients_alchemical_properties.priority' },
+           through: :ingredients_alchemical_properties
+
+  validates :name, presence: true
+  validate :ensure_match_exists
+
+  before_validation :set_canonical_ingredient
+
+  def canonical_ingredients
+    return Canonical::Ingredient.where(id: canonical_ingredient.id) if canonical_ingredient.present?
```
danascheider, line 23: Other non-canonical models would do it like this:
```ruby
return Array.wrap(canonical_ingredient) if canonical_ingredient.present?
```
Why do we need an `ActiveRecord::Relation` here?

app/models/ingredient.rb
```diff
@@ -0,0 +1,55 @@
...
+  def set_canonical_ingredient
+    return if canonical_ingredient.present?
+
+    canonicals = canonical_ingredients
+    self.canonical_ingredient = canonicals.first if canonicals.count == 1
+  end
+
+  def ensure_match_exists
+    return if canonical_ingredient.present?
+    return if Canonical::Ingredient.where('name ILIKE ?', name).present?
```
danascheider, line 51: We could combine these guard clauses as:
```ruby
return if canonical_ingredients.any?
```

app/models/ingredient.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  belongs_to :game
+  belongs_to :canonical_ingredient,
+             class_name: 'Canonical::Ingredient',
+             optional: true,
+             inverse_of: :ingredients
+
+  has_many :ingredients_alchemical_properties, dependent: :destroy, inverse_of: :ingredient
+  has_many :alchemical_properties,
+           -> { select 'alchemical_properties.*, ingredients_alchemical_properties.priority' },
```
danascheider, line 14: We ultimately might want `strength_modifier` and `duration_modifier` attributes to be exposed as well, but since we're not using that right now in the code, we'll save that for when we need it.

spec/models/ingredients_alchemical_property_spec.rb
```diff
@@ -0,0 +1,323 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe IngredientsAlchemicalProperty, type: :model do
+  describe 'validations' do
+    let(:ingredient) { build(:ingredient) }
+
+    before do
+      create(:canonical_ingredient)
```
danascheider, line 10: This object isn't used in all specs. Can we move the `before` block into a context or two so this object isn't created every time?

spec/models/ingredients_alchemical_property_spec.rb
```diff
@@ -0,0 +1,323 @@
...
+      it 'must be greater than zero' do
+        model = build(:ingredients_alchemical_property, duration_modifier: 0)
+
+        model.validate
+        expect(model.errors[:duration_modifier]).to include 'must be greater than 0'
+      end
+    end
+
+    describe 'alchemical_property_id' do
+      it 'must be unique per ingredient' do
```
danascheider, line 157: There should be a spec testing that this doesn't have to be globally unique.

spec/models/ingredients_alchemical_property_spec.rb
```diff
@@ -0,0 +1,323 @@
...
+    end
+
+    context 'when there are no matching canonical models' do
+      subject(:canonical_model) { model.canonical_model }
+
+      let!(:canonical_ingredient) { create(:canonical_ingredient) }
+      let(:ingredient) { create(:ingredient, canonical_ingredient:) }
+      let(:model) { build(:ingredients_alchemical_property, ingredient:) }
+
+      it 'is empty' do
```
danascheider, line 288: Strictly speaking, it's `nil`.

app/models/ingredient.rb
danascheider: Answer: because we call `canonical_ingredients.ids` in the `IngredientsAlchemicalProperty` class and that method is only available on an `ActiveRecord::Relation`.

PR #180, opened 2023-08-06
Create non-canonical MiscItem model

danascheider:
## Context

[**Create non-canonical MiscItem model**](https://trello.com/c/OaJu3oSB/308-create-non-canonical-miscitem-model)

We are creating in-game item models to correspond to each canonical model that already exists in SIM. The next model to be added is the `MiscItem`.

## Changes

* Migration to add `misc_items` table
* `MiscItem` model class
* Specs for `MiscItem` class
* Developer docs

### Required Changes

## Considerations

Many of the considerations pertaining to this model can be found in the included docs. The biggest challenge with this model was that not all `Canonical::MiscItem` records can be uniquely identified by attributes visible to players, so SIM has no way of knowing which canonical model _actually_ matches the item a player has in their inventory. I resolved this problem by assigning ambiguous matches essentially randomly. Matches are narrowed down as much as possible by `name` and `unit_weight`. Then, if there are still multiple matches, the possible matches are tested to see if they are unique items and, if so, if there is already a corresponding in-game item for the given game. Unique canonical models should only have one non-canonical association per game. If all matching models are unique and already have associated models in the same game, a validation error is added to the `MiscItem` indicating it is a duplicate of a unique item.

### DIFF
--- a/app/models/misc_item.rb
+++ b/app/models/misc_item.rb
@@ -9,8 +9,15 @@ class MiscItem < ApplicationRecord
 
   validates :name, presence: true
   validates :unit_weight, numericality: { greater_than_or_equal_to: 0, allow_nil: true }
+  validate :validate_association
+
+  before_validation :set_canonical_misc_item
+
+  DUPLICATE_MESSAGE = 'iks a duplicate of a unique in-game item'
+  DOES_NOT_MATCH = "doesn't match any item that exists in Skyrim"
 
   def canonical_models
+    return [] if name.blank?
     return [canonical_misc_item] if canonical_misc_item.present?
 
     canonicals = Canonical::MiscItem.where('name ILIKE ?', name)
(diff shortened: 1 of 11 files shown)


app/models/misc_item.rb
```diff
@@ -0,0 +1,62 @@
...
+  DUPLICATE_MESSAGE = 'is a duplicate of a unique in-game item'
+  DOES_NOT_MATCH = "doesn't match any item that exists in Skyrim"
+
+  def canonical_models
+    return [] if name.blank?
+    return [canonical_misc_item] if canonical_misc_item.present?
+
+    canonicals = Canonical::MiscItem.where('name ILIKE ?', name)
+
+    attributes_to_match.any? ? canonicals.where(**attributes_to_match) : canonicals
```
danascheider, line 25: This is kind of overkill when we could do:
```ruby
unit_weight.present? ? canonicals.where(unit_weight:) : canonicals
```

I chose not to do this, though, in order to make the API consistent with other canonical models for easier refactoring when we get to [this card](https://trello.com/c/EIdSrLs6/317-revisit-conditional-assignment-of-canonical-models).

spec/models/misc_item_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+        it 'associates the first matching canonical model' do
+          item.validate
+          expect(item.canonical_misc_item).to eq matching_canonicals.first
+        end
+      end
+    end
+
+    context 'when there is already a canonical_misc_item assigned' do
+      let(:canonical_misc_item) { create(:canonical_misc_item, unique_item: true, rare_item: true) }
+      let(:item) { create(:misc_item, canonical_misc_item:) }
```
danascheider, line 245: This should be:
```ruby
build(:misc_item, canonical_misc_item:)
```

PR #182, opened 2023-09-24
Create non-canonical Potion and PotionsAlchemicalProperty models

danascheider:
## Context

[**Create non-canonical Potion model**](https://trello.com/c/LwqylXPf/309-create-non-canonical-potion-model)

We are adding non-canonical versions of each canonical model to prepare for completion of the [inventory list epic](https://trello.com/c/K243CUgb/106-inventory-lists). This PR adds a `Potion` model, corresponding to `Canonical::Potion`. Since each potion can also have alchemical properties, I've also added a `PotionsAlchemicalProperty` join model.

## Changes

* Create non-canonical `Potion` model
* Create non-canonical `PotionsAlchemicalProperty` join model
* Add specs for new models
* Add docs

### Required Changes

## Considerations

Potions are different to some in-game items in that they may or may not have a canonical association. This is because user-created potions can have alchemical properties with nearly infinite combinations of `strength` and `duration` attributes, which depend on factors such as ingredients used, character level, and perks.

I haven't added validations to the `PotionsAlchemicalProperty` model to ensure the presence of a matching canonical model for the same reason (along with a desire to avoid expensive database queries).

A comment on the card for this work points out that there would be no real way to tell if `NULL` values for the `strength` and `duration` of an alchemical property should be matched or not. However, I've decided to match all `NULL` values and treat it as a requirement that the user enter the `strength` and `duration` attributes for the potion when they create it. I believe this will be consistent with a majority of use cases.


app/models/alchemical_property.rb
```diff
@@ -6,13 +6,25 @@ class AlchemicalProperty < ApplicationRecord
...
+  has_many :canonical_ingredients,
+           through: :canonical_ingredients_alchemical_properties,
+           source: :ingredient
 
   has_many :canonical_potions_alchemical_properties,
            dependent: :destroy,
            class_name: 'Canonical::PotionsAlchemicalProperty'
   has_many :canonical_potions, through: :canonical_potions_alchemical_properties, source: :potion
 
+  has_many :ingredients_alchemical_properties,
```
danascheider, line 18: Apparently I neglected to add this code when I created the non-canonical `Ingredient` model 😅

app/models/ingredients_alchemical_property.rb
```diff
@@ -36,15 +36,9 @@ class IngredientsAlchemicalProperty < ApplicationRecord
...
 
   def canonical_models
-    matching_attrs = {
-      alchemical_property_id:,
-      ingredient_id: canonical_ingredients.ids,
-      strength_modifier:,
-      duration_modifier:,
-      priority:,
-    }.compact
+    return Canonical::IngredientsAlchemicalProperty.where(**attributes_to_match) if attributes_to_match.any?
```
danascheider, line 39: This is to create a more uniform API with other canonical models to prepare to generalise these into a single parent class. Not strictly related to this card.

app/models/ingredients_alchemical_property.rb
```diff
@@ -81,4 +75,14 @@ def ensure_match_exists
 
     errors.add(:base, DOES_NOT_MATCH)
   end
+
+  def attributes_to_match
+    {
+      alchemical_property_id:,
+      ingredient_id: canonical_ingredients.ids,
+      strength_modifier:,
```
danascheider, line 83: Should we still match these if they are `nil`?

app/models/potion.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+class Potion < ApplicationRecord
+  belongs_to :game
+  belongs_to :canonical_potion, optional: true, class_name: 'Canonical::Potion'
+
+  has_many :potions_alchemical_properties, dependent: :destroy, inverse_of: :potion
+  has_many :alchemical_properties, through: :potions_alchemical_properties
```
danascheider, line 8: Do we want to scope this like it's scoped in the canonical model, i.e., such that we can use code like this?

```ruby
potion.alchemical_properties.first.strength
potion.alchemical_properties.last.duration
```

app/models/potion.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  validates :name, presence: true
+  validates :unit_weight, numericality: { greater_than_or_equal_to: 0, allow_blank: true }
+
+  before_validation :set_canonical_potion
+
+  def canonical_models
+    return Canonical::Potion.where(id: canonical_potion_id) if canonical_potion.present?
```
danascheider, line 16: The goal of this is to ensure that an `ActiveRecord::Relation` object is always returned from this method and not a single `Canonical::Potion` model. This simplifies the API.

app/models/potion.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    matching = Canonical::Potion.where('name ILIKE ?', name)
+    matching = matching.where('magical_effects ILIKE ?', magical_effects) if magical_effects.present?
+    matching = matching.where(**attributes_to_match) if attributes_to_match.any?
```
danascheider, line 20: This is intended to provide consistency with other in-game item classes. However, the only `attributes_to_match` value is `unit_weight`, which is 0.5 for all potions. This could make for an extraneous database query. Since this method is different from the corresponding method in other in-game item classes anyway (due to the need to match alchemical properties), would it make sense to just remove this line?

app/models/potion.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  private
+
+  def attributes_to_match
+    { unit_weight: }.compact
+  end
+
+  def set_canonical_potion
+    return unless canonical_models.count == 1
+    return if canonical_potion.present?
```
danascheider, line 41: Is this necessary? If the `canonical_potion` is present then the `canonical_models` method will return an `ActiveRecord::Relation` containing only that model. Therefore, this method will already have returned by the time this line runs.

spec/models/potion_spec.rb
```diff
@@ -0,0 +1,412 @@
...
+            name: 'potion of healing',
+          )
+        end
+
+        before do
+          create(:canonical_potion)
+        end
+
+        it 'matches case-insensitively' do
+          expect(canonical_models).to eq matching_canonicals
```
danascheider, line 60: Let's mirror the above test and use `contain_exactly`.

spec/models/potion_spec.rb
```diff
@@ -0,0 +1,412 @@
...
+            unit_weight: 0.5,
+          )
+        end
+
+        before do
+          create(:canonical_potion, name: 'potion of healing', unit_weight: 0.6)
+        end
+
+        it 'returns all matching canonicals' do
+          expect(canonical_models).to eq matching_canonicals
```
danascheider, line 81: This should also use `contain_exactly`.

spec/models/potion_spec.rb
```diff
@@ -0,0 +1,412 @@
...
+            magical_effects: 'Foobar',
+          )
+        end
+
+        before do
+          create(:canonical_potion, name: 'potion of healing', magical_effects: nil)
+        end
+
+        it 'returns all matching canonicals' do
+          expect(canonical_models).to eq matching_canonicals
```
danascheider, line 102: Use `contain_exactly`.

app/models/potion.rb
danascheider: So far we don't need this so let's not use default scopes when there's no reason to.

app/models/potion.rb
danascheider: Note: there are tests that rely on `unit_weight` being matched.

app/models/ingredients_alchemical_property.rb
danascheider: Answer: no. In fact, we probably shouldn't even define these methods on the `IngredientsAlchemicalProperty` itself and should instead delegate the `strength_modifier` and `duration_modifier` methods to the `Canonical::IngredientsAlchemicalProperty` model.

danascheider: Update: I have [created a card](https://trello.com/c/Qxnqde8s/326-remove-strengthmodifier-and-durationmodifier-from-non-canonical-ingredientsalchemicalproperty-model) for this.

PR #186, opened 2023-09-25
Create non-canonical Property model

danascheider:
## Context

[**Create non-canonical Property model**](https://trello.com/c/bPRqA9hP/310-create-non-canonical-property-model)

Although there turns out to already be a non-canonical `Property` model in our code base and database schema, it needed some work to align it with the `Canonical::Property` model, which just added several new fields in [this PR](https://github.com/danascheider/skyrim_inventory_management/pull/185). Additionally, the class was not in line with the design of the other in-game item models, in that it didn't find its own canonical model or set values from that model like the others.

## Changes

* Add fields to `properties` table corresponding to new fields on `canonical_properties`
* Add validations for the new fields
* Add `HomesteadValidator` class to validate mutually exclusive fields for [homestead properties](https://elderscrolls.fandom.com/wiki/Homestead_(Hearthfire))
* Make `Property` models identify and assign their own `Canonical::Property`
* Set fields on `Property` from selected `Canonical::Property`
* Add/update tests for new class and behaviour
* Add docs on the `Property` class
* Fix typo in README

### Required Changes

## Considerations

Because there is such a limited set of canonical properties and all are uniquely named, I've decided to make the `Property` model identify the canonical model based on `name` alone (whereas to be truly in line with other non-canonical models, it should match on `name`, `city`, and `hold`). The result is that, if there is a canonical model with the same `name` (case-insensitively), the `city` and `hold` values, if they differ from those on the canonical model, are replaced by the values on the canonical model. The `name` is also changed to that on the canonical model in order to keep casing consistent.

I wasn't sure what validations should be the responsibility of the `HomesteadValidator` and which should be the responsibility of validations in the `Property` class itself. I opted to put all validations pertaining to homestead-only fields in the `HomesteadValidator`. The exceptions are `has_apiary`, `has_grain_mill` and `has_fish_hatchery`. Each of these fields can only be true for one specific property, each of which is a homestead. However, since they apply to individual properties and not to homesteads generally, I decided to put those validations in the `Property` class.


app/models/property.rb
```diff
@@ -25,15 +25,39 @@ class Property < ApplicationRecord
 
   validates :city,
             inclusion: { in: Canonical::Property::VALID_CITIES, message: 'must be a Skyrim city in which an ownable property is located', allow_blank: true },
-            uniqueness: { scope: :game_id, message: 'must be unique per game if present', allow_blank: true }
+            uniqueness: { scope: :game_id, message: 'must be unique per game if present', allow_nil: true }
+
+  validate :ensure_alchemy_lab_available, if: -> { has_alchemy_lab == true && canonical_property&.alchemy_lab_available == false }
```
danascheider, line 30: We don't want this validation to fail if the `canonical_property` is `nil`, hence why we can't use `has_alchemy_lab == true && !canonical_property&.alchemy_lab_available`.

app/models/property.rb
```diff
@@ -25,15 +25,38 @@ class Property < ApplicationRecord
...
+  DOES_NOT_MATCH = "doesn't match any ownable property that exists in Skyrim"
+
   private
 
+  def set_canonical_model
+    self.canonical_property ||= Canonical::Property.find_by('name ILIKE ?', name)
+  end
+
+  def set_values_from_canonical
+    return if canonical_property.nil?
```
danascheider, line 53: This case will be dealt with at validation.

PR #189, opened 2023-09-28
Create non-canonical Staff model

danascheider:
## Context

[**Create non-canonical Staff model**](https://trello.com/c/HRIc59Zc/311-create-non-canonical-staff-model)

We are creating non-canonical versions of each (most) of the canonical models in the database. The next one up is the `Staff` model, corresponding to `Canonical::Staff`.

## Changes

* Migration to add `staves` table
* New `Staff` model
* Add and update factories
* Tests for the new class
* Docs for `Canonical::Staff` and `Staff` models
* Add unenchanted staves to JSON data

### Required Changes

## Considerations

I considered using `delegate` to delegate the `#spells` and `#powers` methods directly to the `Canonical::Staff`, but I wanted an `ActiveRecord::Relation` returned even if this was `nil`, so instead I implemented a method that returns either `Canonical::Staff#spells`/`Canonical::Staff#powers` or, if the canonical staff is `nil`, an empty `ActiveRecord::Relation`. This will have the disadvantage that we won't be able to access the `strength` attribute stored in the `Canonical::StavesSpell` join model, but so far we don't need to anyway.

In keeping with other canonical models, the `Staff`'s `#canonical_models` method will return an `ActiveRecord::Relation` containing only its associated `canonical_staff` if there is already one associated. We have identified that this is an issue as other changes made to attributes of an in-game item could change what the association should be. There is a [card](https://trello.com/c/EIdSrLs6/317-revisit-conditional-assignment-of-canonical-models) to deal with this issue for all in-game item models. There is [another card](https://trello.com/c/KfDVb7ZF/329-ensure-in-game-item-models-continue-to-match-canonicals-after-update) to make sure in-game items are validated against canonicals on update.

### DIFF
--- a/app/models/game.rb
+++ b/app/models/game.rb
@@ -9,6 +9,10 @@ class Game < ApplicationRecord
   has_many :clothing_items, dependent: :destroy
   has_many :ingredients, dependent: :destroy
   has_many :jewelry_items, dependent: :destroy
+  has_many :misc_items, dependent: :destroy
+  has_many :potions, dependent: :destroy
+  has_many :properties, dependent: :destroy
+  has_many :staves, dependent: :destroy
 
   # `before_save` callbacks need to be defined before
   # `before_destroy` callbacks, which need to be defined here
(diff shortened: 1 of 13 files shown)


app/models/game.rb
```diff
@@ -9,6 +9,10 @@ class Game < ApplicationRecord
   has_many :clothing_items, dependent: :destroy
   has_many :ingredients, dependent: :destroy
   has_many :jewelry_items, dependent: :destroy
+  has_many :misc_items, dependent: :destroy
+  has_many :potions, dependent: :destroy
+  has_many :properties, dependent: :destroy
+  has_many :staves, dependent: :destroy
```
danascheider, line 15: Had to define all these associations - I'd neglected to do so when I introduced these models.

PR #211, opened 2023-10-09
Create non-canonical weapon model

danascheider:
## Context

[**Create non-canonical Weapon model**](https://trello.com/c/fVI5kDfy/312-create-non-canonical-weapon-model)

We are creating non-canonical versions of most of our canonical models. The next one is the `Weapon` model. This is one of the most critical in-game item types in Skyrim.

## Changes

* Migration creating `weapons` table
* Non-canonical `Weapon` model
* Validations and hooks to `EnchantablesEnchantment` against canonical models
* Small changes to other canonical models with associations to `EnchantablesEnchantment`
* Tests for the new model
* Docs for the new model

### Required Changes

## Considerations

Testing this work was challenging because the desired start state for a lot of test cases was precluded by SIM's validations and hooks. I ended up having to exploit edge cases to make sure that test data got instantiated correctly. For example, to make sure that weapons have enchantments added from a canonical match after saving, I had to first create multiple canonical models without enchantments, save the weapon (so that there would be multiple canonical matches and no `canonical_weapon` would be assigned), and then add enchantments to the canonicals to ensure they would be added at the appropriate time during the tests.

I decided not to add validations to the `Weapon` model to test for a correct combination of `weapon_type` and `category`, because the `Canonical::Weapon` model already has this validation, meaning that any `Weapon` with an incorrect combination would fail matching against canonicals.

## TODO

- [x] Test that enchantments are matched based on `strength` and not just enchantment ID

### DIFF
--- a/app/models/weapon.rb
+++ b/app/models/weapon.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class Weapon < ApplicationRecord
+  belongs_to :game
+  belongs_to :canonical_weapon,
+             optional: true,
+             class_name: 'Canonical::Weapon',
+             inverse_of: :armors
+end
--- a/app/models/weapon.rb
+++ b/app/models/weapon.rb
@@ -75,6 +75,14 @@ class Weapon < ApplicationRecord
     canonicals.uniq
   end
 
+  def crafting_materials
+    canonical_weapon&.crafting_materials
+  end
+
+  def tempering_materials
+    canonical_weapon&.tempering_materials
+  end
+
   private
 
   def set_canonical_weapon
--- a/app/models/weapon.rb
+++ b/app/models/weapon.rb
@@ -5,5 +5,94 @@ class Weapon < ApplicationRecord
   belongs_to :canonical_weapon,
              optional: true,
              class_name: 'Canonical::Weapon',
-             inverse_of: :armors
+             inverse_of: :weapons
+
+  has_many :enchantables_enchantments,
+           dependent: :destroy,
+           as: :enchantable
+  has_many :enchantments,
+           -> { select 'enchantments.*, enchantables_enchantments.strength as strength' },
+           through: :enchantables_enchantments
+
+  validates :name, presence: true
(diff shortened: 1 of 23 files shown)


app/models/weapon.rb
```diff
@@ -0,0 +1,128 @@
+# frozen_string_literal: true
+
+class Weapon < ApplicationRecord
+  include ActiveModel::Dirty
```
danascheider, line 4: We don't need this as it is already included in `ApplicationRecord` objects. The module only needs to be included in classes that don't inherit from `ActiveRecord::Base`.

spec/support/factories/weapons.rb
```diff
@@ -0,0 +1,25 @@
...
+    game
+
+    name { 'Dwarven War Axe' }
+
+    trait :with_matching_canonical do
+      association :canonical_weapon,
+                  factory: :canonical_weapon,
+                  strategy: :create
+
+      name { canonical_weapon.name }
```
danascheider, line 14: We don't really need to define this here since the default name for a canonical weapon is the same as that for an in-game weapon.

spec/support/factories/weapons.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    trait :with_enchanted_canonical do
+      association :canonical_weapon,
+                  factory: %i[canonical_weapon with_enchantments],
+                  strategy: :create
+
+      name { canonical_weapon.name }
```
danascheider, line 22: Same comment here.

spec/models/weapon_spec.rb
```diff
@@ -0,0 +1,606 @@
...
+    end
+  end
+
+  describe '::before_validation' do
+    subject(:validate) { weapon.validate }
+
+    context 'when there is a canonical model assigned' do
+      let(:canonical_weapon) { weapon.canonical_weapon }
+
+      context 'when the canonical model has no enchantments' do
```
danascheider, line 42: This context is not really required since the next context tests the same things.

PR #216, opened 2023-10-23
Create non-canonical Book model

danascheider:
## Context

[**Create non-canonical Book model**](https://trello.com/c/S4Tgbmse/304-create-non-canonical-book-model)

We are creating in-game items corresponding to each of our canonical models (see docs). The `Book` model is the last model we need to create :tada:

The `Book` model needed an association to either `Ingredient` or `Canonical::Ingredient` due to the fact that books of `book_type` `"recipe"` can have ingredients associated. I chose to make the association to `Canonical::Ingredient` (see "Considerations", below), and achieved this by making the existing `Canonical::RecipesIngredients` model non-canonical and making its relation to `recipe` polymorphic, so that the `recipe` can be either a `Book` or a `Canonical::Book`. 

## Changes

* New non-canonical `Book` model
* Rename `Canonical::RecipesIngredient` to `RecipesCanonicalIngredient` in code, factories, docs, and tests
* Make association to `recipe` from `RecipesCanonicalIngredient` polymorphic (add `recipe_type` field and remove foreign key constraint)
* Ensure that the `recipe_type` for existing `Canonical::RecipesIngredient` models gets set to `Canonical::Book` in the database
* Clean up JSON data for canonical books to remove books that differ only in `item_code` or contents from other books in the database
* Add/update specs for new and changed models
* Update docs

### Required Changes

### Canonical::Ingredient Association

The `Canonical::Book` model has an association to `Canonical::Ingredient` that can be used when the book is a recipe. If I followed the pattern, I would've created a `RecipesIngredient` model defining the relationship between `Book`s and `Ingredient`s. However, a recipe book is not associated to a specific instance of an ingredient - it's associated to the ingredient as a whole. For this reason, I thought it would be more useful to use the `Canonical::Ingredient` for this association. Since the association now can involve a non-canonical model, I changed the name of the join model to move it outside the `Canonical::` namespace.

### JSON Data Changes

Some canonical books in the database are differentiable mainly by contents and none of the attributes we track, other than item code, actually differ. In this PR, I've removed all these books from the database and set `unique_item` to `false` on the single remaining model. It would be better to identify the item based on contents, but especially given the extraneous nature of most books in the game, it seems unlikely users would want to enter even part of their text to disambiguate.


app/models/canonical/book.rb
```diff
@@ -49,6 +54,10 @@ def self.unique_identifier
       :item_code
     end
 
+    def recipe?
```
danascheider, line 57: Add this method to make a somewhat uniform API between possible associations to the `RecipesCanonicalIngredient` model.

app/models/canonical/ingredient.rb
```diff
@@ -17,11 +17,11 @@ class Ingredient < ApplicationRecord
...
              through: :canonical_ingredients_alchemical_properties
 
-    has_many :canonical_recipes_ingredients,
+    has_many :recipes_canonical_ingredients,
              dependent: :destroy,
-             class_name: 'Canonical::RecipesIngredient',
+             class_name: 'RecipesCanonicalIngredient',
              inverse_of: :ingredient
-    has_many :recipes, through: :canonical_recipes_ingredients, class_name: 'Canonical::Book', source: :recipe
+    has_many :recipes, through: :recipes_canonical_ingredients, class_name: 'Canonical::Book', source: :recipe
```
danascheider, line 24: We should not have `class_name` here because it could be `Book` or `Canonical::Book` now.

spec/models/canonical/sync/books_spec.rb
```diff
@@ -176,7 +176,7 @@
 
       it "doesn't destroy associations" do
         perform
-        expect(book_in_json.reload.canonical_ingredients.length).to eq 3
+        expect(book_in_json.reload.canonical_ingredients.length).to eq 5
```
danascheider, line 179: This value changed because of a change we made to the factory.

PR #217, opened 2023-10-28
Add NOT NULL constraint to name/title attributes on in-game items

danascheider:
## Context

[**Add NOT NULL constraint to name attribute for all in-game items**](https://trello.com/c/R9ouYfHL/321-add-not-null-constraint-to-name-attribute-for-all-in-game-items)

In-game items are matched with canonical models first using their `name` attribute (`title` for books). There are validations for the presence of the `name`/`title` on each model, but there are some missing `NOT NULL` constraints on the database. We also would like to make sure that the `unit_weight` attribute of all canonical and in-game items has `precision: 5, scale: 2` set.

## Changes

* Migration to ensure a `NOT NULL` constraint has been added to the `name`/`title` attribute of each in-game item model
* Migration to ensure `scale: 2` and `precision: 5` on the `unit_weight` of all canonical and in-game items
* Schema updates

### Required Changes

### DIFF
--- a/db/migrate/20231028000539_add_not_null_constraint_for_name_attribute_of_in_game_items.rb
+++ b/db/migrate/20231028000539_add_not_null_constraint_for_name_attribute_of_in_game_items.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class AddNotNullConstraintForNameAttributeOfInGameItems < ActiveRecord::Migration[7.1]
+  def change
+    change_column_null :armors, :name, false
+    change_column_null :books, :title, false
+    change_column_null :clothing_items, :name, false
+    change_column_null :ingredients, :name, false
+    change_column_null :jewelry_items, :name, false
+    change_column_null :misc_items, :name, false
+    change_column_null :potions, :name, false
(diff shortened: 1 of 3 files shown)


db/migrate/20231028000539_add_not_null_constraint_for_name_attribute_of_in_game_items.rb
```diff
@@ -0,0 +1,16 @@
...
+    change_column_null :books, :title, false
+    change_column_null :clothing_items, :name, false
+    change_column_null :ingredients, :name, false
+    change_column_null :jewelry_items, :name, false
+    change_column_null :misc_items, :name, false
+    change_column_null :potions, :name, false
+    change_column_null :properties, :name, false
+    change_column_null :staves, :name, false
+    change_column_null :weapons, :name, false
+  end
```
danascheider, line 15: An astute reviewer will notice that there are more changes in the migration file than in the `schema.rb` file. However, when I removed extraneous lines from this file and ran the migration again, `NOT NULL` constraints began to be _removed_ from the schema, so I added them back here.

PR #221, opened 2023-10-30
Prevent unique canonicals from having multiple associations in the same game

danascheider:
## Context

[**Prevent unique canonicals from having multiple associations in the same game**](https://trello.com/c/NzCjnIVC/328-prevent-unique-canonicals-from-having-multiple-associations-in-the-same-game)

Most of our canonical models have a boolean field called `unique_item`. If this field is set to `true`, there is only one instance of the item in the game. We want to prevent users from creating multiple instances of unique items in SIM.

## Changes

* Prevent users from creating multiples of `unique_item`s for the same `game` in SIM
* Add tests for new behaviour
* Update docs
* Miscellaneous minor cleanup/refactors

### Required Changes

## Considerations

There is still the troubling issue of unique items that respawn. Respawning items should really not be considered unique. There is a [card](https://trello.com/c/x5bBR50K/342-dont-consider-respawning-items-unique) to change this.

### DIFF
--- a/app/validators/armor_validator.rb
+++ b/app/validators/armor_validator.rb
@@ -3,6 +3,7 @@
 class ArmorValidator < ActiveModel::Validator
   NO_CANONICAL_MATCHES = "doesn't match an armor item that exists in Skyrim"
   DOES_NOT_MATCH = 'does not match value on canonical model'
+  DUPLICATE_MATCH = 'is a duplicate of a unique in-game item'
 
   def validate(record)
     @record = record
@@ -22,8 +23,21 @@ class ArmorValidator < ActiveModel::Validator
   def validate_against_canonical
     canonical_armor = record.canonical_armor
 
+    validate_unique_canonical_match if canonical_armor.unique_item
+
     record.errors.add(:unit_weight, DOES_NOT_MATCH) unless record.unit_weight == canonical_armor.unit_weight
     record.errors.add(:weight, DOES_NOT_MATCH) unless record.weight == canonical_armor.weight
     record.errors.add(:magical_effects, DOES_NOT_MATCH) unless record.magical_effects == canonical_armor.magical_effects
+  end
+
+  def validate_unique_canonical_match
+    return unless record.canonical_armor.unique_item
+
+    armors = record.canonical_armor.armors.where(game_id: record.game_id)
+
+    return if armors.count < 1
+    return if armors.count == 1 && armors.first == record
+
+    record.errors.add(:base, DUPLICATE_MATCH)
   end
 end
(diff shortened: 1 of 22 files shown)


app/validators/armor_validator.rb
```diff
@@ -22,8 +23,21 @@ def validate(record)
   def validate_against_canonical
     canonical_armor = record.canonical_armor
 
+    validate_unique_canonical_match if canonical_armor.unique_item
```
danascheider, line 26: We could remove this redundant conditional.

app/validators/armor_validator.rb
```diff
@@ -22,8 +23,21 @@ def validate(record)
...
+
     record.errors.add(:unit_weight, DOES_NOT_MATCH) unless record.unit_weight == canonical_armor.unit_weight
     record.errors.add(:weight, DOES_NOT_MATCH) unless record.weight == canonical_armor.weight
     record.errors.add(:magical_effects, DOES_NOT_MATCH) unless record.magical_effects == canonical_armor.magical_effects
   end
+
+  def validate_unique_canonical_match
+    return unless record.canonical_armor.unique_item
```
danascheider, line 34: We should use the safe_navigation operator: `record.canonical_armor&.unique_item`.

spec/models/book_spec.rb
```diff
@@ -21,6 +21,72 @@
...
+          create(
+            :canonical_book,
+            unique_item: true,
+            rare_item: true,
+          )
+        end
+
+        context 'when the canonical model has no books' do
+          before do
+            create(:book, canonical_book:)
```
danascheider, line 58: We shouldn't do this in a context called `'when the canonical model has no books'`.

spec/models/book_spec.rb
```diff
@@ -21,6 +21,72 @@
...
+          end
+
+          it 'is valid' do
+            validate
+            expect(book.errors[:base]).to be_blank
```
danascheider, line 63: Let's use `be_empty` since that's used elsewhere for checking arrays. We could also just use `expect(book).to be_valid`.

spec/models/canonical/armor_spec.rb
```diff
@@ -38,7 +38,7 @@
         expect(armor.errors[:weight]).to include "can't be blank"
       end
 
-      it 'is invalid with an invalid weight value' do
+      it 'is invalid with an invalid unit_weight value' do
```
danascheider, line 41: We need to change this back, it is supposed to be `weight`.

spec/models/ingredient_spec.rb
```diff
@@ -50,10 +52,76 @@
...
+            :canonical_ingredient,
+            unique_item: true,
+            rare_item: true,
+          )
+        end
+
+        context 'when the canonical ingredient has no other matches' do
+          it 'is valid' do
+            validate
+            expect(ingredient.errors[:base]).to be_empty
```
danascheider, line 94: Let's see if we can use `expect(ingredient).to be_valid` here.

PR #222, opened 2023-11-01
Move potion type designation from Canonical::Potion to AlchemicalProperty

danascheider:
## Context

[**Remove potion_type field from Canonical::Potion**](https://trello.com/c/JPj2WxPA/322-remove-potiontype-field-from-canonicalpotion)

The `Canonical::Potion` model has a field called `potion_type` designating whether the potion is a potion or a poison. However, this is not great because, unlike most other in-game items, not all `Potion` models ultimately correspond to a `Canonical::Potion` as many are user-created. Consequently, it would be impossible to determine the potion type of user-created potions. By moving the `potion_type` value to an `effect_type` field on the `AlchemicalProperty` model, we enable all potions with alchemical properties to be identified as potions, poisons, or mixed effects.

## Changes

* Migration to add `effect_type` field to `AlchemicalProperty` model
* Tests and validations for `effect_type` on `AlchemicalProperty`
* Migration to remove `potion_type` from `Canonical::Potion` model
* Remove tests and validations for `potion_type`
* Update canonical data so alchemical properties have an `effect_type` and there is no `potion_type` for canonical potions

### Required Changes

## Considerations

We will want a `NOT NULL` constraint on the `effect_type` column, however, this could not be added in this PR because it will need to be added after the models in the database have been backfilled.


db/migrate/20231031213740_add_effect_type_to_alchemical_properties.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddEffectTypeToAlchemicalProperties < ActiveRecord::Migration[7.1]
+  def change
+    add_column :alchemical_properties, :effect_type, :string
```
danascheider, line 5: As mentioned in "Considerations", we ultimately want this column `NOT NULL`, but have to backfill data before we can do that.

PR #226, opened 2023-11-03
Ensure no more than 4 alchemical effects per potion

danascheider:
## Context

[**Ensure no more than 4 alchemical properties per potion**](https://trello.com/c/fjIrjlQb/325-ensure-no-more-than-4-alchemical-properties-per-potion)

Potions in Skyrim, whether user-created or backed by a canonical model, cannot have more than 4 alchemical effects. Canonical potions also never have more than 4 effects in the unmodded game.  We want to add validations to make sure that these models can't be associated to more than 4 alchemical properties.

## Changes

* Add validation to `Canonical::PotionsAlchemicalProperty` ensuring a max of 4 can be created per potion
* Add validation to `PotionsAlchemicalProperty` ensuring a max of 4 can be created per potion
* Test new validations
* Add line to docs

### Required Changes

### DIFF
--- a/app/models/potions_alchemical_property.rb
+++ b/app/models/potions_alchemical_property.rb
@@ -21,4 +21,22 @@ class PotionsAlchemicalProperty < ApplicationRecord
               only_integer: true,
               allow_nil: true,
             }
+
+  validate :ensure_max_per_potion
+
+  MAX_PER_POTION = Canonical::PotionsAlchemicalProperty::MAX_PER_POTION
+
+  private
+
+  def ensure_max_per_potion
+    return if potion.alchemical_properties.length < MAX_PER_POTION
+    return if persisted? &&
+      !potion_id_changed? &&
+      potion.alchemical_properties.length == MAX_PER_POTION
+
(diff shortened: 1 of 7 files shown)


app/models/potions_alchemical_property.rb
```diff
@@ -21,4 +21,22 @@ class PotionsAlchemicalProperty < ApplicationRecord
               only_integer: true,
               allow_nil: true,
             }
+
+  validate :ensure_max_per_potion
+
+  MAX_PER_POTION = Canonical::PotionsAlchemicalProperty::MAX_PER_POTION
```
danascheider, line 27: Let's keep our line length reasonable here

spec/models/potion_spec.rb
```diff
@@ -195,6 +195,8 @@
               strength: nil,
               duration: nil,
             )
+
+            potion.reload
```
danascheider, line 199: These needed to be here to begin with - the tests were wrong and it wasn't until they started failing due to this change that I realised it.

PR #230, opened 2023-11-05
Rematch canonical ingredient if changed attributes lead to a different match

danascheider:
## Context

When an `Ingredient` model is associated to a `Canonical::Ingredient`, calling `#canonical_ingredients` on the model returns the `Canonical::Ingredient` model that is already assigned, in an `ActiveRecord::Relation` object. The goal of this was to avoid expensive database lookups when there was already an assigned canonical model. However, the result is that, if an ingredient's attributes change, it results in validation errors since it is validated against the existing canonical - even if there is another canonical that matches. We want the database to be queried again if the `Ingredient` no longer matches its assigned canonical.

## Changes

* Change `#canonical_ingredients` method to be `#canonical_models` in line with other in-game item classes
* Enable associated canonical ingredient to be changed if attributes are updated
* Update `IngredientsAlchemicalProperty` model to use `ingredient.canonical_models` instead of delegating `#canonical_ingredients` to `ingredient`
* Update tests

### Required Changes

## Related PRs

* #227 
* #228 
* #229

### DIFF
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -22,23 +22,27 @@ class Ingredient < ApplicationRecord
   DOES_NOT_MATCH = "doesn't match an ingredient that exists in Skyrim"
   DUPLICATE_MATCH = 'is a duplicate of a unique in-game item'
 
-  def canonical_ingredients
-    return Canonical::Ingredient.where(id: canonical_ingredient.id) if canonical_ingredient.present?
+  def canonical_model
+    canonical_ingredient
+  end
 
-    matching = Canonical::Ingredient.where('name ILIKE ?', name)
-    matching = matching.where(**attributes_to_match) if attributes_to_match.any?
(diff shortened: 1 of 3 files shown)


app/models/ingredient.rb
```diff
@@ -22,31 +22,35 @@ class Ingredient < ApplicationRecord
...
+  def canonical_model
+    canonical_ingredient
+  end
+
+  def canonical_models
+    return Canonical::Ingredient.where(id: canonical_ingredient.id) if canonical_model_matches?
 
-    matching = Canonical::Ingredient.where('name ILIKE ?', name)
-    matching = matching.where(**attributes_to_match) if attributes_to_match.any?
+    canonicals = Canonical::Ingredient.where('name ILIKE ?', name)
```
danascheider, line 32: Changed these names to be more similar to other in-game item classes.

PR #233, opened 2023-11-06
Rematch canonical potion if changed attributes lead to a different match

danascheider:
## Context

Currently, if a potion is associated to a `Canonical::Potion`, it remains so for all time and its attributes get overwritten by the attributes from that canonical. However, the behaviour we want is slightly different. If the attributes of a potion have changed, we want it to check if the associated `Canonical::Potion` still matches, if another `Canonical::Potion` matches now, or if no canonical potion matches. If there is another matching canonical, it should be set as the `canonical_potion`.

## Changes

* Generate new `Canonical::Potion` matches if changed attributes lead to another match being better
* Clear `Canonical::Potion` association if changed attributes lead to missing or non-unique matches
* Update tests

### Required Changes

## Related PRs

* #227 
* #228 
* #229 
* #230 
* #231 
* #232

### DIFF
--- a/app/models/potion.rb
+++ b/app/models/potion.rb
@@ -16,15 +16,19 @@ class Potion < ApplicationRecord
 
   DUPLICATE_MATCH = 'is a duplicate of a unique in-game item'
 
+  def canonical_model
+    canonical_potion
+  end
+
   def canonical_models
-    return Canonical::Potion.where(id: canonical_potion_id) if canonical_potion.present?
+    return Canonical::Potion.where(id: canonical_potion_id) if canonical_model_matches?
 
-    matching = Canonical::Potion.where('name ILIKE ?', name)
-    matching = matching.where('magical_effects ILIKE ?', magical_effects) if magical_effects.present?
+    canonicals = Canonical::Potion.where('name ILIKE ?', name)
+    canonicals = canonicals.where('magical_effects ILIKE ?', magical_effects) if magical_effects.present?
 
-    return matching if matching.blank? || alchemical_properties.none?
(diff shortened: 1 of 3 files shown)


app/models/potion.rb
```diff
@@ -16,15 +16,19 @@ class Potion < ApplicationRecord
 
   DUPLICATE_MATCH = 'is a duplicate of a unique in-game item'
 
+  def canonical_model
```
danascheider, line 19: Let's add a test for this method.

app/models/potion.rb
```diff
@@ -16,15 +16,19 @@ class Potion < ApplicationRecord
...
+    canonical_potion
+  end
+
   def canonical_models
-    return Canonical::Potion.where(id: canonical_potion_id) if canonical_potion.present?
+    return Canonical::Potion.where(id: canonical_potion_id) if canonical_model_matches?
 
-    matching = Canonical::Potion.where('name ILIKE ?', name)
-    matching = matching.where('magical_effects ILIKE ?', magical_effects) if magical_effects.present?
+    canonicals = Canonical::Potion.where('name ILIKE ?', name)
```
danascheider, line 26: Change variable name for consistency with this method in other in-game item classes.

spec/support/factories/potions.rb
```diff
@@ -4,7 +4,7 @@
   factory :potion do
     game
 
-    name { 'Deadly Poison' }
+    name { 'My Potion' }
```
danascheider, line 7: In general, we want the default name/title for factories of non-canonical models to match the default name/title for the corresponding canonical model. In this case, that name is `'My Potion'`.

PR #234, opened 2023-11-07
Rematch canonical property if changed attributes lead to a different match

danascheider:
## Context

Currently, once a `Property` model is associated to a `Canonical::Property`, the association doesn't change. However, if attributes change, it would be better UX to simply change the associated canonical rather than raising validation errors, raising validation errors only if the change results in no canonical matches. This PR ensures that the canonical model can change.

## Changes

* Enable `Canonical::Property` association to change on the `Property` model if attributes change
* Update tests
* Add `game` association to `property` factory
* Fix (unrelated) flaky test

### Required Changes

## Considerations

Unlike other models, the `Property` model is required to have a `canonical_property` association. This is because canonical properties are uniquely named and limited in number, so there would be no scenario in which a match was ambiguous.

## Related PRs

* #227 
* #228 
* #229 
* #230 
* #231 
* #232 
* #233

### DIFF
--- a/docs/in_game_items/property.md
+++ b/docs/in_game_items/property.md
@@ -1,10 +1,10 @@
 # Property
 
-The `Property` class represents in-game instances of the [`Canonical::Property`](/docs/canonical_models/canonical-property.md) type. Properties differ slightly from other in-game items in two ways:
+The `Property` class represents in-game instances of the [`Canonical::Property`](/docs/canonical_models/canonical-property.md) type. Properties differ slightly from other in-game items in three ways:
 
 * A property is also a location (in some cases with sublocations)
 * A property can be associated with inventory and shopping lists (functionality that will be built out further for the MVP)
-
+* All properties must always be associated to a `Canonical::Property`
 
 Additionally, a number of characteristics of a property cannot be determined by the `Canonical::Property` associated. Users can add different amenities or, in the case of [homesteads](https://elderscrolls.fandom.com/wiki/Homestead_(Hearthfire)), build entirely different rooms on the property. For that reason, the `Canonical::Property` model has a number of attributes like `alchemy_lab_available` and `enchanters_tower_available` indicating whether the amenity or room _can be added_ to the property instead of whether it is actually present.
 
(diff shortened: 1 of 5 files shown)


docs/in_game_items/property.md
```diff
@@ -5,11 +5,14 @@ The `Property` class represents in-game instances of the [`Canonical::Property`]
 * A property is also a location (in some cases with sublocations)
 * A property can be associated with inventory and shopping lists (functionality that will be built out further for the MVP)
 
+
```
danascheider, line 8: Oopsie, extra line.

spec/models/canonical/sync/ingredients_spec.rb
```diff
@@ -101,7 +101,7 @@
         it 'adds alchemical properties' do
           perform
           expect(item_in_json.alchemical_properties.pluck(:name))
-            .to eq ['Weakness to Frost', 'Fortify Sneak', 'Weakness to Poison', 'Fortify Restoration']
+            .to contain_exactly('Weakness to Frost', 'Fortify Sneak', 'Weakness to Poison', 'Fortify Restoration')
```
danascheider, line 104: It turns out these are not deterministically ordered in order of priority. However, this test is not testing the priority order, it is only testing that the associations are created. It's better for that purpose to test that the array contains the right elements, rather than that the order is the same.

spec/models/property_spec.rb
```diff
@@ -222,28 +206,25 @@
 
     describe 'grain mill availability' do
       context 'when a grain mill is not available at the given property' do
-        let(:canonical_property) { Canonical::Property.find_by(name: 'Lakeview Manor') }
+        let(:property) { build(:property) }
```
danascheider, line 209: The default name is `"Lakeview Manor"`

spec/support/factories/properties.rb
```diff
@@ -2,10 +2,8 @@
 
 FactoryBot.define do
   factory :property do
-    name { 'Lakeview Manor' }
+    game
 
-    trait :with_matching_canonical do
```
danascheider, line 7: This isn't being used anywhere because the `Property` tests actually sync canonicals, preventing more from being created, and it doesn't seem to be used elsewhere either.

PR #235, opened 2023-11-07
Rematch canonical staff if changed attributes lead to a different match

danascheider:
## Context

Currently, once a `Staff` model is associated to a `Canonical::Staff` (i.e., its `canonical_staff_id` has been set), the canonical staff associated no longer changes. Instead, any attribute changes the user makes get overwritten on the next `before_validation` hook. This is bad UX if it would be possible to instead change the `canonical_staff` to one that matches the new attributes. In this PR, the canonical staff is changed in this scenario, or the association removed if the changes result in an ambiguous (or no) match. As before, if there are no matching canonicals at all, validation will fail.

## Changes

* Allow reassignment of `canonical_staff` on `Staff` model when attributes change
* Set `canonical_staff` to `nil` if changed attributes result in an ambiguous (or no) canonical match
* Update tests
* Reformat code in `Spell` class to eliminate long lines

### Required Changes

## Related PRs

* #227 
* #228 
* #229 
* #230 
* #231 
* #232 
* #233 
* #234

### DIFF
--- a/app/models/staff.rb
+++ b/app/models/staff.rb
@@ -27,35 +27,36 @@ class Staff < ApplicationRecord
     canonical_staff&.powers || Power.none
   end
 
+  def canonical_model
+    canonical_staff
+  end
+
   def canonical_models
-    return Canonical::Staff.where(id: canonical_staff.id) if canonical_staff.present?
+    return Canonical::Staff.where(id: canonical_staff_id) if canonical_model_matches?
 
-    canonicals = Canonical::Staff.where('name ILIKE ?', name)
+    query = 'name ILIKE :name'
+    query += ' AND (magical_effects ILIKE :magical_effects)' unless magical_effects.nil?
 
+    canonicals = Canonical::Staff.where(query, name:, magical_effects:)
     attributes_to_match.any? ? canonicals.where(**attributes_to_match) : canonicals
   end
 
   private
 
   def set_canonical_staff
-    return if canonical_staff.present?
-    return unless canonical_models.count == 1
+    unless canonical_models.count == 1
+      clear_canonical_staff
+      return
+    end
 
-    canonical = canonical_models.first
-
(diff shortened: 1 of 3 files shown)


app/models/staff.rb
```diff
@@ -27,35 +27,36 @@ def powers
...
+    canonical_staff
+  end
+
   def canonical_models
-    return Canonical::Staff.where(id: canonical_staff.id) if canonical_staff.present?
+    return Canonical::Staff.where(id: canonical_staff_id) if canonical_model_matches?
 
-    canonicals = Canonical::Staff.where('name ILIKE ?', name)
+    query = 'name ILIKE :name'
+    query += ' AND (magical_effects ILIKE :magical_effects)' unless magical_effects.nil?
```
danascheider, line 38: Since `magical_effects` can be lengthy, it'd be good to not require users to capitalise perfectly.

app/models/staff.rb
```diff
@@ -27,35 +27,36 @@ def powers
...
-    return if canonical_staff.present?
-    return unless canonical_models.count == 1
-
-    canonical = canonical_models.first
+    unless canonical_models.count == 1
+      clear_canonical_staff
+      return
+    end
 
-    return if canonical.unique_item && canonical.staves.where(game_id:).any?
```
danascheider, line 46: This is now handled in validations.

app/models/staff.rb
```diff
@@ -27,35 +27,36 @@ def powers
...
     self.unit_weight = canonical_staff.unit_weight
     self.magical_effects = canonical_staff.magical_effects
   end
 
   def attributes_to_match
-    {
-      unit_weight:,
-      magical_effects:,
-    }.compact
+    { unit_weight: }.compact
```
danascheider, line 59: Since `magical_effects` are now matched case insensitively.

PR #237, opened 2023-11-07
Nullify canonical models that no longer match

danascheider:
## Context

During the course of our work enabling canonical matches to be changed if an in-game item is updated, we realised that, for many of the prior PRs related to this, we failed to ensure that canonical models that no longer match get removed as associations (unless being immediately replaced by one that matches better). This could cause problems if an update to an in-game item model resulted in an ambiguous match with multiple canonical models (or none for that matter) but the `canonical_<model>` association remained.

While working on this PR, I unearthed some serious bugs. While in previous PRs we ensured that the `#canonical_models` method for each model returned the correct canonical match(es), but in the case of several models we failed to test, or implement, the changes that would actually cause the correct match to be _set_ as the `canonical_<model>`. These bugs are fixed in this PR, as are minor logic errors around `magical_effects` fields, which are now matched case-insensitively with canonical models.

## Changes

* Ensure canonical associations are being set correctly on in-game items
* Nullify the `canonical_<model>` field if an attribute update leads to ambiguous, or no, canonical matches
* Fix bug where `nil` `magical_effects` value leads to canonical models never matching
* Ensure `magical_effects` fields are always matched case-insensitively to canonicals
* Some cleanup and performance optimisations
* Add missing tests
* Test new behaviour

### Required Changes

## Related PRs

* #227 
* #228 
* #229 
* #230 
* #231 
* #232 
* #233 
* #234 
* #235 
* #236

### DIFF
--- a/app/models/armor.rb
+++ b/app/models/armor.rb
@@ -42,7 +42,7 @@ class Armor < ApplicationRecord
   end
 
   def canonical_models
-    return Canonical::Armor.where(id: canonical_armor.id) if canonical_model_matches?
+    return Canonical::Armor.where(id: canonical_armor_id) if canonical_model_matches?
 
     query = 'name ILIKE :name'
     query += ' AND magical_effects ILIKE :magical_effects' if magical_effects.present?
--- a/app/models/armor.rb
+++ b/app/models/armor.rb
@@ -44,7 +44,10 @@ class Armor < ApplicationRecord
   def canonical_models
     return Canonical::Armor.where(id: canonical_armor.id) if canonical_model_matches?
 
-    canonicals = Canonical::Armor.where('name ILIKE ?', name)
+    query = 'name ILIKE :name'
+    query += ' AND magical_effects ILIKE :magical_effects' if magical_effects.present?
+
+    canonicals = Canonical::Armor.where(query, name:, magical_effects:)
     attributes_to_match.any? ? canonicals.where(**attributes_to_match) : canonicals
   end
 
@@ -59,13 +62,16 @@ class Armor < ApplicationRecord
   private
 
   def set_canonical_armor
-    return unless canonical_models.count == 1
+    unless canonical_models.count == 1
+      clear_canonical_armor
+      return
+    end
 
     self.canonical_armor = canonical_models.first
     self.name = canonical_armor.name # in case casing differs
+    self.magical_effects = canonical_armor.magical_effects
     self.unit_weight = canonical_armor.unit_weight
     self.weight = canonical_armor.weight
-    self.magical_effects = canonical_armor.magical_effects
 
     set_enchantments if persisted?
   end
@@ -81,6 +87,10 @@ class Armor < ApplicationRecord
     end
   end
 
+  def clear_canonical_armor
+    self.canonical_armor_id = nil
+  end
+
   def canonical_model_matches?
     return false if canonical_model.nil?
     return false unless name.casecmp(canonical_model.name).zero?
@@ -94,7 +104,6 @@ class Armor < ApplicationRecord
     {
       unit_weight:,
       weight:,
-      magical_effects:,
     }.compact
   end
 end
--- a/app/models/armor.rb
+++ b/app/models/armor.rb
@@ -62,12 +62,14 @@ class Armor < ApplicationRecord
   private
 
   def set_canonical_armor
-    unless canonical_models.count == 1
+    canonicals = canonical_models
+
+    unless canonicals.count == 1
       clear_canonical_armor
       return
     end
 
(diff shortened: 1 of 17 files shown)


app/models/armor.rb
```diff
@@ -59,13 +62,18 @@ def tempering_materials
   private
 
   def set_canonical_armor
-    return unless canonical_models.count == 1
+    canonicals = canonical_models
```
danascheider, line 65: The `#canonical_models` method will always do database calls, which can be expensive. For that reason, we should only call it once here. I've made the same change in the other classes.

app/validators/clothing_item_validator.rb
```diff
@@ -2,13 +2,12 @@
 
 class ClothingItemValidator < ActiveModel::Validator
   NO_CANONICAL_MATCHES = "doesn't match a clothing item that exists in Skyrim"
-  DOES_NOT_MATCH = 'does not match value on canonical model'
```
danascheider, line 5: Remove unused constant.

app/validators/clothing_item_validator.rb
```diff
@@ -2,13 +2,12 @@
...
   DUPLICATE_MATCH = 'is a duplicate of a unique in-game item'
 
   def validate(record)
     @record = record
 
-    if @record.canonical_models.blank?
+    if @record.canonical_models.none?
```
danascheider, line 10: This doesn't actually change behaviour but I thought it was clearer.

spec/models/armor_spec.rb
```diff
@@ -88,43 +88,7 @@
   end
 
   describe '::before_validation' do
-    context 'when there is a single matching canonical model' do
```
danascheider, line 91: We should really not remove this context.

spec/models/armor_spec.rb
```diff
@@ -156,10 +120,85 @@
...
+          validate
+
+          expect(armor.name).to eq 'Imperial Boots of Resist Frost'
+          expect(armor.magical_effects).to eq 'This Will Be Case Insensitive'
+          expect(armor.weight).to eq 'light armor'
+          expect(armor.unit_weight).to eq 2
+        end
+      end
+
+      context 'when the update results in an ambiguous match' do
```
danascheider, line 168: Let's add a test to this context to ensure attributes are not set from ambiguous canonical matches.

spec/models/book_spec.rb
```diff
@@ -265,5 +265,78 @@
...
+      context 'when the update results in no canonical matches' do
+        it 'sets the canonical_book to nil' do
+          book.title = "sinderion's field journal"
+
+          expect { validate }
+            .to change(book, :canonical_book)
+                  .to(nil)
+        end
+
+        it "doesn't update attributes" do
```
danascheider, line 332: We don't really need this test because, if there aren't any canonical matches, then what would it update attributes from anyway?

spec/models/ingredient_spec.rb
```diff
@@ -280,16 +282,90 @@
...
+
+          expect(ingredient.name).to eq 'horseradish'
+          expect(ingredient.unit_weight).to be_nil
+        end
+      end
+
+      context 'when the update results in no canonical matches' do
+        it 'sets the canonical ingredient association to nil' do
+          ingredient.name = 'horseradish'
+          ingredient.unit_weight = nil
```
danascheider, line 362: It's a small thing, but changing the name would be enough to result in no canonical matches - this is just one more operation that doesn't affect the test outcome.

spec/models/jewelry_item_spec.rb
```diff
@@ -299,9 +301,84 @@
...
+
+          expect(item.name).to eq 'silver jeweled necklace'
+          expect(item.unit_weight).to be_nil
+        end
+      end
+
+      context 'when the update results in no canonical matches' do
+        it 'sets the canonical jewelry item to nil' do
+          item.name = 'silver jeweled necklace'
+          item.unit_weight = nil
```
danascheider, line 374: Same here - really don't need to change both attributes to result in no canonical matches.

spec/models/misc_item_spec.rb
```diff
@@ -251,106 +253,81 @@
...
       end
     end
 
-    context "when multiple complete matches can't be further differentiated" do
-      let(:game) { create(:game) }
-      let(:item) { build(:misc_item, name: 'Skull', unit_weight: 2, game:) }
+    context 'when updating in-game item attributes' do
+      let(:item) { create(:misc_item, :with_matching_canonical) }
 
-      context 'when a canonical model indicates a unique item' do
```
danascheider, line 263: Don't worry, this functionality is still tested elsewhere in the file. These tests were ~duplicates.

PR #238, opened 2023-11-10
Match armor models to canonicals based on enchantments

danascheider:
## Context

[**Match Armor models to canonicals based on enchantments**](https://trello.com/c/sv9Nc1Mw/347-match-armor-models-to-canonicals-based-on-enchantments)

When an in-game item model is able to have enchantments, alchemical properties, or other associations, those associations should be used along with attributes to match the model to possible canonical matches. However, we seem to have overlooked this on the Armor model.

## Changes

* Ensure that enchantments on `Armor` models are considered, along with attributes, in identifying canonical matches
* Add tests for new behaviour
* Update docs
* Remove some contexts from the weapon spec that were the only contexts at their level of nesting

### Required Changes

## Considerations

There is [another card](https://trello.com/c/9i7Z0bjc/345-ensure-associations-updated-when-canonical-models-rematched) to ensure that, if the canonical model changes, associations are added or removed as appropriate.

I'll also check our other models and make sure that others with associations are also matched based on those associations.

### DIFF
--- a/app/models/armor.rb
+++ b/app/models/armor.rb
@@ -48,7 +48,28 @@ class Armor < ApplicationRecord
     query += ' AND magical_effects ILIKE :magical_effects' if magical_effects.present?
 
     canonicals = Canonical::Armor.where(query, name:, magical_effects:)
-    attributes_to_match.any? ? canonicals.where(**attributes_to_match) : canonicals
+    canonicals = attributes_to_match.any? ? canonicals.where(**attributes_to_match) : canonicals
+
+    return canonicals if enchantments.none?
+
+    enchantables_enchantments.each do |join_model|
+      canonicals = if join_model.strength.present?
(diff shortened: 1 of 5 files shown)


app/models/armor.rb
```diff
@@ -48,7 +48,28 @@ def canonical_models
     query += ' AND magical_effects ILIKE :magical_effects' if magical_effects.present?
 
     canonicals = Canonical::Armor.where(query, name:, magical_effects:)
-    attributes_to_match.any? ? canonicals.where(**attributes_to_match) : canonicals
+    canonicals = attributes_to_match.any? ? canonicals.where(**attributes_to_match) : canonicals
```
danascheider, line 51: We could just do:
```ruby
canonicals = canonicals.where(**attributes_to_match) if attributes_to_match.any?
```

PR #239, opened 2023-11-10
Match Canonical::ClothingItem models based on enchantments

danascheider:
## Context

[**Match ClothingItem models to canonicals based on enchantments**](https://trello.com/c/He2sZUBs/348-match-clothingitem-models-to-canonicals-based-on-enchantments)

If an in-game item is enchantable, any enchantments it has should be taken into consideration when setting a canonical match. However, this logic is missing from a couple of our models, including `ClothingItem` and [`JewelryItem`](https://trello.com/c/lGhtcHKf/349-match-jewelryitem-models-to-canonicals-based-on-enchantments). (The logic was added to `Armor` in #238.)

In the course of doing this work, I noticed that matching logic included queries like,

```ruby
canonicals.left_outer_joins(:enchantables_enchantments).where(
  '(enchantables_enchantments.enchantment_id = :enchantment_id AND enchantables_enchantments.enchantable_type = :type AND enchantables_enchantments.strength = :strength) OR canonical_clothing_items.enchantable = true',
  enchantment_id: join_model.enchantment_id,
  type: 'Canonical::ClothingItem',
  strength: join_model.strength,
)
```

The observant reader will notice that `AND enchantables_enchantments.enchantable_type = :type` is redundant here, because we are already using a join with a `Canonical::ClothingItem` relation. No `EnchantablesEnchantment` in this relation will have any other `enchantable_type`. Since this issue was present in other models too, I've removed it from them as well.

## Changes

* Consider enchantments when matching `ClothingItem` models to canonical matches
* Remove redundant `AND enchantables_enchantments.enchantable_type = :type` from queries on `ClothingItem`, `Armor`, and `Weapon` models
* Add tests for new behaviour
* Update docs

### Required Changes

## Considerations

There is additional [planned work](https://trello.com/c/9i7Z0bjc/345-ensure-associations-updated-when-canonical-models-rematched) to update/remove associations when a canonical model changes or is removed.

### DIFF
--- a/app/models/armor.rb
+++ b/app/models/armor.rb
@@ -55,16 +55,14 @@ class Armor < ApplicationRecord
     enchantables_enchantments.each do |join_model|
       canonicals = if join_model.strength.present?
                      canonicals.left_outer_joins(:enchantables_enchantments).where(
-                       '(enchantables_enchantments.enchantment_id = :enchantment_id AND enchantables_enchantments.enchantable_type = :type AND enchantables_enchantments.strength = :strength) OR canonical_armors.enchantable = true',
+                       '(enchantables_enchantments.enchantment_id = :enchantment_id AND enchantables_enchantments.strength = :strength) OR canonical_armors.enchantable = true',
                        enchantment_id: join_model.enchantment_id,
-                       type: 'Canonical::Armor',
                        strength: join_model.strength,
                      )
                    else
                      canonicals.left_outer_joins(:enchantables_enchantments).where(
-                       '(enchantables_enchantments.enchantment_id = :enchantment_id AND enchantables_enchantments.enchantable_type = :type) OR canonical_armors.enchantable = true',
+                       '(enchantables_enchantments.enchantment_id = :enchantment_id) OR canonical_armors.enchantable = true',
                        enchantment_id: join_model.enchantment_id,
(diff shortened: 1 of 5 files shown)


app/models/armor.rb
```diff
@@ -55,16 +55,14 @@ def canonical_models
...
-                       '(enchantables_enchantments.enchantment_id = :enchantment_id AND enchantables_enchantments.enchantable_type = :type AND enchantables_enchantments.strength = :strength) OR canonical_armors.enchantable = true',
+                       '(enchantables_enchantments.enchantment_id = :enchantment_id AND enchantables_enchantments.strength = :strength) OR canonical_armors.enchantable = true',
                        enchantment_id: join_model.enchantment_id,
-                       type: 'Canonical::Armor',
                        strength: join_model.strength,
                      )
                    else
                      canonicals.left_outer_joins(:enchantables_enchantments).where(
-                       '(enchantables_enchantments.enchantment_id = :enchantment_id AND enchantables_enchantments.enchantable_type = :type) OR canonical_armors.enchantable = true',
+                       '(enchantables_enchantments.enchantment_id = :enchantment_id) OR canonical_armors.enchantable = true',
```
danascheider, line 64: We need to check that `strength` is `NULL` in this case.

PR #243, opened 2023-11-19
Designate enchantments and alchemical properties added automatically

danascheider:
## Context

Certain in-game item models, when matched with a corresponding canonical model, can have any associations from those canonical models that don't currently exist on the in-game model added when the model is saved. For example, if a `Weapon` is matched with a `Canonical::Weapon` that has enchantments, any of the canonical weapon's enchantments that are missing from the in-game weapon will be added.

Models with this feature include:

* `Armor` (`Enchantment`)
* `ClothingItem` (`Enchantment`)
* `JewelryItem` (`Enchantment`)
* `Potion` (`AlchemicalProperty`)
* `Weapon` (`Enchantment`)

(Note that this PR also adds functionality to add alchemical properties to potions automatically, a feature that was missing.)

We have recently also added the capability for a canonical model to be changed (if a new one matches) or removed (if an update to the in-game item results in multiple possible canonical matches). However, these automatically-added associations are not removed or changed. Since canonical models are matched based on these associations in addition to attributes, this can prevent a new canonical model from being accurately identified. We want to remove associations added from a canonical model when that canonical model is removed as the canonical model for a given in-game item.

The fly in the ointment is that the user might've added some of the associations themselves. We assume that users do this based on knowledge they have about the item, so these associations should not be removed when a canonical model is changed. In order to differentiate, we needed to add an `added_automatically` attribute to the join models for these associations, which will be set to `true` when the association is added automatically and `false` otherwise. There will also be an `added_automatically` scope on each of these join models, which will enable us to more easily target them for removal when a canonical match changes.

## Changes

* Add boolean `added_automatically` column to `EnchantablesEnchantment` and `PotionsAlchemicalProperty` models
* Add `added_automatically` scope to join models to identify those added automatically
* Add alchemical properties from canonical potions when a non-canonical potion is saved
* Set `added_automatically` to `true` when a join model is created automatically based on a canonical model
* Set `added_automatically` to `false` otherwise
* Add tests for the above functionality
* Update docs

### Required Changes

## Considerations

Note that this PR does not yet remove these associations when the canonical model changes. It only provides a mechanism for doing so.

### DIFF
--- a/app/models/armor.rb
+++ b/app/models/armor.rb
@@ -104,7 +104,9 @@ class Armor < ApplicationRecord
       enchantables_enchantments.find_or_create_by!(
         enchantment_id: model.enchantment_id,
         strength: model.strength,
-      )
+      ) do |created_model|
+        created_model.added_automatically = true
+      end
     end
   end
 
--- a/app/models/armor.rb
+++ b/app/models/armor.rb
@@ -104,9 +104,7 @@ class Armor < ApplicationRecord
       enchantables_enchantments.find_or_create_by!(
         enchantment_id: model.enchantment_id,
         strength: model.strength,
-      ) do |created_model|
-        created_model.added_automatically = true
-      end
+      ) {|new_model| new_model.added_automatically = true }
     end
   end
 
--- a/db/migrate/20231111205621_add_added_automatically_to_join_models.rb
+++ b/db/migrate/20231111205621_add_added_automatically_to_join_models.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class AddAddedAutomaticallyToJoinModels < ActiveRecord::Migration[7.1]
+  def up
+    add_column :enchantables_enchantments,
+               :added_automatically,
+               :boolean,
+               default: false
+
+    add_column :potions_alchemical_properties,
+               :added_automatically,
+               :boolean,
+               default: false
+
+    # rubocop:disable Rails/SkipsModelValidations
+    EnchantablesEnchantment.update_all(added_automatically: true)
+    PotionsAlchemicalProperty.update_all(added_automatically: true)
+    # rubocop:enable Rails/SkipsModelValidations
+
+    change_column_null :enchantables_enchantments, :added_automatically, false
+    change_column_null :potions_alchemical_properties, :added_automatically, false
+  end
+
+  def down
+    remove_column :enchantables_enchantments, :added_automatically
+    remove_column :potions_alchemical_properties, :added_automatically
+  end
+end
--- a/spec/models/armor_spec.rb
+++ b/spec/models/armor_spec.rb
@@ -287,6 +287,12 @@ RSpec.describe Armor, type: :model do
           expect(armor.enchantments.length).to eq 2
         end
 
+        it 'sets "added_automatically" to true on new associations' do
+          armor.save!
+
+          expect(armor.enchantables_enchantments.pluck(:added_automatically).uniq).to eq [true]
+        end
+
         it 'sets the correct strengths', :aggregate_failures do
           armor.save!
           matching_canonical.enchantables_enchantments.each do |join_model|
@@ -311,6 +317,11 @@ RSpec.describe Armor, type: :model do
 
         it "doesn't remove the existing enchantments" do
           expect(armor.enchantments.reload.length).to eq 4
+        end
+
+        it 'sets "added_automatically" only on the new associations' do
+          expect(armor.enchantables_enchantments.pluck(:added_automatically))
+            .to eq [true, true, false, false]
         end
       end
     end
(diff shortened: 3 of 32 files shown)


app/models/armor.rb
```diff
@@ -104,7 +104,7 @@ def set_enchantments
       enchantables_enchantments.find_or_create_by!(
         enchantment_id: model.enchantment_id,
         strength: model.strength,
-      )
+      ) {|new_model| new_model.added_automatically = true }
```
danascheider, line 107: This is a little-known feature of `find_or_create_by!` where a block can be passed to add attributes to a created model that are not used to find an existing model.

db/migrate/20231111205621_add_added_automatically_to_join_models.rb
```diff
@@ -0,0 +1,28 @@
...
+               default: false
+
+    add_column :potions_alchemical_properties,
+               :added_automatically,
+               :boolean,
+               default: false
+
+    # rubocop:disable Rails/SkipsModelValidations
+    EnchantablesEnchantment.update_all(added_automatically: true)
+    PotionsAlchemicalProperty.update_all(added_automatically: true)
```
danascheider, line 17: It doesn't really matter what we set these to since there are no `PotionsAlchemicalProperty` models in the prod database right now. For `EnchantablesEnchantment` it should be set to `true` since all of these models correspond to automatically-created canonical enchantables.

spec/models/armor_spec.rb
```diff
@@ -287,6 +287,12 @@
           expect(armor.enchantments.length).to eq 2
         end
 
+        it 'sets "added_automatically" to true on new associations' do
+          armor.save!
+
+          expect(armor.enchantables_enchantments.pluck(:added_automatically).uniq).to eq [true]
```
danascheider, line 293:
```suggestion
      expect(armor.enchantables_enchantments.pluck(:added_automatically))
        .to be_all(true)
```

spec/models/clothing_item_spec.rb
```diff
@@ -202,6 +202,12 @@
           expect(item.enchantments.length).to eq 2
         end
 
+        it 'sets "added_automatically" to true on new associations' do
+          item.save!
+
+          expect(item.enchantables_enchantments.pluck(:added_automatically).uniq).to eq [true]
```
danascheider, line 208: Same suggestion as above

spec/models/jewelry_item_spec.rb
```diff
@@ -451,16 +451,66 @@
...
+        end
+
+        it 'adds enchantments from the canonical model' do
+          item.save!
+          expect(item.enchantments.length).to eq 2
+        end
+
+        it 'sets "added_automatically" to true on new associations' do
+          item.save!
+          expect(item.enchantables_enchantments.pluck(:added_automatically).uniq)
```
danascheider, line 477: Same suggestion as above

spec/models/potion_spec.rb
```diff
@@ -550,4 +550,71 @@
...
+      it 'adds alchemical properties from the canonical potion' do
+        potion.save!
+
+        expect(potion.alchemical_properties.reload.length).to eq 2
+      end
+
+      it 'sets "added_automatically" to true on new associations' do
+        potion.save!
+
+        expect(potion.potions_alchemical_properties.pluck(:added_automatically).uniq).to eq [true]
```
danascheider, line 575: Same suggestion as above

spec/models/weapon_spec.rb
```diff
@@ -379,110 +379,61 @@
...
-            )
-          end
+        it 'sets "added_automatically" to true on new associations' do
+          weapon.save!
 
-          it 'adds enchantments from the canonical' do
-            add_enchantment
-            expect(weapon.reload.enchantments.length).to eq 2
-          end
+          expect(weapon.enchantables_enchantments.pluck(:added_automatically).uniq).to eq [true]
```
danascheider, line 406: Same suggestion as above

spec/models/armor_spec.rb
```diff
@@ -312,6 +318,11 @@
         it "doesn't remove the existing enchantments" do
           expect(armor.enchantments.reload.length).to eq 4
         end
+
+        it 'sets "added_automatically" only on the new associations' do
+          expect(armor.enchantables_enchantments.pluck(:added_automatically))
+            .to eq [true, true, false, false]
```
danascheider, line 324: Because of how the tests are set up, the automatically added ones will be first.

spec/models/jewelry_item_spec.rb
```diff
@@ -451,16 +451,66 @@
...
+            .to eq [true]
+        end
+
+        it 'sets the correct strengths', :aggregate_failures do
+          item.save!
+
+          matching_canonical.enchantables_enchantments.each do |join_model|
+            has_matching = item.enchantables_enchantments.any? do |model|
+              model.enchantment == join_model.enchantment && model.strength == join_model.strength
```
danascheider, line 486: Currently this will be be `nil` - we should adjust either the factory or the test setup to make sure there is actually a value here.

PR #246, opened 2023-11-20
Update factory_bot_rails to 6.4.0 from 6.2.0

danascheider:
## Context

In #244 Dependbot tried to update the factory_bot_rails package. However, in that PR, the transitive dependency on `factory_bot` is locked at v6.4.0 and it seems to be causing weird failures in the database step of GitHub Actions. In this PR, the transitive dependency is locked at v6.4.1 and hopefully that's a patch that fixes this issue.

## Changes

* Update `factory_bot_rails` from 6.2.0 to 6.4.0
* Ensure `factory_bot` dependency is at 6.4.1

### Required Changes

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,7 +37,7 @@ group :development, :test do
   gem 'database_cleaner-active_record', '~> 2.1'
 
   # Use FactoryBot to create models for tests
-  gem 'factory_bot_rails', '~> 6.2.0'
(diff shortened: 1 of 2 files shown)


danascheider:
This issue was caused by [this known bug](https://github.com/thoughtbot/factory_bot_rails/issues/433). A fix is pending and will hopefully be released later.

PR #250, opened 2023-11-23
Only match canonicals using manually added associations

danascheider:
## Context

When an enchantable in-game item or a potion is matched with a canonical model, the matching algorithm takes into account the enchantments of the item or alchemical properties of the potion. If a unique match is identified, any enchantments or alchemical properties the canonical match has that are missing from the in-game item are added. In #243 we added the ability to identify which enchantments or alchemical properties were added manually by the user and which were added automatically from a canonical model. The ultimate goal is to enable automatically added associations to be removed if the associated canonical no longer matches (since having them still there could prevent another accurate match from being made, or eliminate all possible matches).

The first step to achieve this is to make sure that the matching algorithm only considers associations added by the user when identifying a canonical match, since automatically-added associations should be details filled in after identifying a canonical match, not used to infer the canonical match. This PR ensures that only enchantments or alchemical properties manually added by the user are considered in the matching algorithm. 

## Changes

* Ensure matching algorithms for enchantable items only take into account manually added enchantments
* Ensure matching algorithm for potions only take into account manually added alchemical properties
* Tests for the above
* Add some missing test coverage
* Update docs

### Required Changes


spec/models/weapon_spec.rb
```diff
@@ -92,6 +92,148 @@
     end
   end
 
+  describe '#canonical_model' do
+    subject(:canonical_model) { weapon.canonical_model }
+
+    context 'when there is a canonical armor associated' do
```
danascheider, line 98: This should say `'when there is a canonical weapon associated'`.

PR #252, opened 2023-11-24
Ensure associations updated when canonical models rematched

danascheider:
## Context

Currently, although automatically added `EnchantablesEnchantment` and `PotionsAlchemicalProperty` associations are not considered in matching canonical models to enchantable items or potions (respectively), these associations remain when the canonical model changes or the association set to `nil`. When a canonical model changes, these associations should be removed and replaced with the new canonical model's enchantments or alchemical properties, and if there is no clear matching canonical, these associations should just be removed. Enchantments or alchemical properties added manually by the user will remain.

## Changes

* Ensure enchantable items have automatically-added enchantments removed/updated when the canonical model changes or is removed
* Ensure potions have automatically-added alchemical properties removed/updated when the canonical potion changes or is removed
* Fix deprecation warning regarding `fixture_path`, which is now a `fixture_paths` array in Rails 7.1
* Add tests for new behaviour
* Update docs

### Required Changes

## Considerations

Re-matching canonical models when an `EnchantablesEnchantment` or `PotionsAlchemicalProperty` model was added automatically resulted in a stack-level error. For that reason, and since automatically added join models aren't considered for canonical matching anyway, it was necessary to only update the canonical enchantable or canonical potion when the join model created was added manually.


app/models/armor.rb
```diff
@@ -94,12 +94,14 @@ def set_canonical_armor
...
     self.weight = canonical_armor.weight
 
-    set_enchantments if persisted?
+    set_enchantments if persisted? && canonical_armor_id_changed?
   end
 
   def set_enchantments
     return if canonical_armor.enchantments.empty?
 
+    enchantables_enchantments.added_automatically.find_each(&:destroy!)
```
danascheider, line 103: Should we create a method for this since this code occurs twice (in each of these in-game models)?

app/models/clothing_item.rb
```diff
@@ -77,12 +77,14 @@ def set_canonical_clothing_item
...
+    set_enchantments if persisted? && canonical_clothing_item_id_changed?
   end
 
   def set_enchantments
     return if canonical_clothing_item.enchantments.empty?
 
+    enchantables_enchantments.added_automatically.find_each(&:destroy!)
+
     canonical_clothing_item.enchantables_enchantments.each do |model|
       enchantables_enchantments.find_or_create_by!(
```
danascheider, line 89: We might not even need to use `find_or_create_by!` given that this will only run if the canonical model has changed. Let's give it a try locally and see what tests break 🤠

danascheider: Update: tests did break, and the reason they did is that the user-added enchantment, _of course,_ could and probably will be one of the same ones already present on a canonical. So never mind on this one.

spec/rails_helper.rb
```diff
@@ -45,7 +45,7 @@
   config.include AuthHelper
 
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
-  config.fixture_path = Rails.root.join('spec', 'fixtures')
+  config.fixture_paths = [Rails.root.join('spec', 'fixtures')]
```
danascheider, line 48: This is to fix a deprecation warning.

spec/models/armor_spec.rb
```diff
@@ -246,6 +282,19 @@
...
         end
+
+        it 'removes automatically added enchantments', :aggregate_failures do
+          armor.name = 'imperial boots of resist frost'
+          armor.magical_effects = 'this will be case insensitive'
+          armor.weight = nil
+          armor.unit_weight = nil
+
+          validate
+          armor.reload
```
danascheider, line 293: Not that it particularly matters for this test, but this will change the values back to the ones in the database. We should use `armor.enchantables_enchantments.reload` instead.

PR #254, opened 2023-11-25
Introduce superclass for enchantable in-game items

danascheider:
## Context

[**Review in-game item classes to ensure consistency**](https://trello.com/c/KXA86ONy/324-review-in-game-item-classes-to-ensure-consistency)

We want our in-game item models to be as consistent as possible to improve maintainability and reduce the likelihood of bugs in the already-complicated process of matching and rematching canonical models. A review of the in-game item models revealed some differences that would make it difficult for all to inherit from a single parent class. However, four classes - `Armor`, `ClothingItem`, `JewelryItem`, and `Weapon` - had enough similarities to extract an abstract parent class for them.

We will continue to work on reducing duplication and improving consistency in other models but this PR is large enough so I've decided to make those changes in another PR.

Other than adding missing validations for unique canonical models, this PR does not change application behaviour.

## Changes

* Create `EnchantableInGameItem` abstract superclass, inheriting from `ApplicationRecord`
* Make `Armor`, `ClothingItem`, `JewelryItem`, and `Weapon` inherit from this superclass
* Add validations for unique canonical models where they are missing
* Ensure tests pass

### Required Changes

## Considerations

I considered an `ActiveSupport::Concern` for this, but it required a bunch of things to be defined before the module could be included, and that was ugly so I decided to use a superclass instead.


app/models/armor.rb
```diff
@@ -80,39 +35,38 @@ def tempering_materials
...
+  def canonical_class
+    Canonical::Armor
+  end
 
-    unless canonicals.count == 1
-      clear_canonical_armor
-      return
-    end
+  def canonical_table
+    'canonical_armors'
```
danascheider, line 43: This could have been extracted as a constant, but this way we can define this in the superclass with a `NotImplementedError` and more easily find if this method isn't defined.

app/models/clothing_item.rb
```diff
@@ -1,81 +1,49 @@
...
-      return
-    end
+  def inverse_relationship_name
+    :clothing_items
+  end
+
+  alias_method :canonical_model=, :canonical_clothing_item=
+
+  def set_values_from_canonical
+    return if canonical_model.nil?
```
danascheider, line 42: We should really be c checking here if the canonical model has changed.

app/models/weapon.rb
```diff
@@ -87,52 +45,39 @@ def tempering_materials
...
-    return if weapons.count < 1
-    return if weapons.count == 1 && weapons.first == self
+    self.name = canonical_model.name
+    self.unit_weight = canonical_model.unit_weight
+    self.category = canonical_model.category
+    self.weapon_type = canonical_model.weapon_type
+    self.magical_effects = canonical_model.magical_effects
 
-    errors.add(:base, DUPLICATE_MATCH)
+    set_enchantments if persisted? && canonical_model_id_changed?
```
danascheider, line 80: `canonical_model_id_changed?` is redundant here because of the guard clause.

app/models/enchantable_in_game_item.rb
```diff
@@ -0,0 +1,142 @@
...
+  validate :validate_unique_canonical
+  validate :ensure_canonicals_exist
+
+  before_validation :set_canonical_model
+  before_validation :set_values_from_canonical
+
+  after_create :set_enchantments, if: -> { canonical_model.present? }
+
+  def canonical_model
+    raise NotImplementedError.new("#{MUST_DEFINE} public \#canonical_model method")
```
danascheider, line 29: I don't think we actually need the escape character in the error message.

PR #262, opened 2023-11-28
Correct errors in canonical misc item data

danascheider:
## Context

[**Don't consider respawning items unique**](https://trello.com/c/x5bBR50K/342-dont-consider-respawning-items-unique)

Because SIM prevents unique items from being created multiple times in the same game, it's critical that all items designated as canonically unique actually be unique and not, under any circumstances, obtainable multiple times. Oftentimes, supposedly "unique" items are, in fact, obtainable multiple times, either through glitches or respawning. Because of the large number of canonical models with `unique_item` set to `true`, we've decided to make these changes on a one-model-per-PR basis. This PR handles misc items. However, there is a change to one book that was discovered to be non-unique during the course of researching the misc items.

## Changes

* Correct inaccurate data on canonical misc items, especially around uniqueness
* Mark the book "Glover's Letter" as non-unique since it respawns

### DIFF
--- a/lib/tasks/canonical_models/canonical_misc_items.json
+++ b/lib/tasks/canonical_models/canonical_misc_items.json
@@ -9,7 +9,7 @@
       ],
       "description": null,
       "purchasable": false,
-      "unique_item": true,
+      "unique_item": false,
       "rare_item": true,
       "quest_item": true,
       "quest_reward": true
@@ -153,7 +153,7 @@
       ],
       "description": null,
       "purchasable": false,
-      "unique_item": true,
+      "unique_item": false,
       "rare_item": true,
       "quest_item": false,
       "quest_reward": false
@@ -185,7 +185,7 @@
       ],
       "description": null,
       "purchasable": false,
-      "unique_item": true,
+      "unique_item": false,
       "rare_item": true,
       "quest_item": true,
       "quest_reward": false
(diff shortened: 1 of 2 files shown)


lib/tasks/canonical_models/canonical_misc_items.json
```diff
@@ -1387,39 +1387,7 @@
       ],
       "description": null,
       "purchasable": false,
-      "unique_item": true,
-      "rare_item": true,
-      "quest_item": true,
-      "quest_reward": false
-    }
-  },
-  {
```
danascheider, line 1396: I had to delete two duplicates of the Opaque Vessel and the Werewolf Totem, since they were differentiable only by item code, and set the third duplicate to `unique_item: false`. Having the three copies of these models would've prevented a canonical model from ever being set on an in-game item with any of these as the canonical.

lib/tasks/canonical_models/canonical_misc_items.json
```diff
@@ -1787,15 +1723,15 @@
...
+      "unique_item": false,
       "rare_item": true,
       "quest_item": true,
       "quest_reward": false
     }
   },
   {
     "attributes": {
-      "name": "Silver Hand Strategem",
+      "name": "Silver Hand Stratagem",
```
danascheider, line 1734: Turns out it's canonically misspelled.

PR #310, opened 2024-03-12
Change name of Canonical::Material to Canonical::RawMaterial

danascheider:
## Context

We are adding a new model to account for the fact that "materials" used for smithing and building are not necessarily actual `Canonical::Material`s. We wanted a way to classify what we are now calling `Canonical::RawMaterial`s such as ores and ingots separately from the `Material`s ultimately used to make a thing at a forge. As a first step, we are changing the name of the `canonical_materials` table to `canonical_raw_materials`, as well as updating names of classes, factories, fixtures, associations, etc.

## Changes

* Rename `canonical_materials` table to `canonical_raw_materials`
* Rename `Canonical::Material` model to `Canonical::RawMaterial`
* Update syncers and Rake tasks
* Rename JSON data files
* Update references, tests and docs

### Required Changes

## Considerations

Although I changed the class name to `RawMaterial`, I've kept the name of the `material` association the same for the `Canonical::TemperablesTemperingMaterial` and `Canonical::CraftablesCraftingMaterial` model. This is because, as part of work on the same card, we will be introducing a new `Canonical::Material` model that is used for this association. This model will provide a uniform interface for materials that may not be actual raw materials (for example, creation of a dwarven crossbow requires an actual crossbow as a material, but users will have likely classified this as a weapon in their inventory lists).


docs/canonical_models/syncing-canonical-models.md
```diff
@@ -94,4 +94,4 @@ Syncers provide detailed info and error logging. The logging is done using the R
 
 ## Curating JSON Data
 
-Information about the JSON data, its structure, generating and curating it is available in the [docs on canonical data](/docs/canonical_models/canonical-data.md).
\ No newline at end of file
+Information about the JSON data, its structure, generating and curating it is available in the [docs on canonical data](/docs/canonical_models/canonical-data.md).
```
danascheider, line 97: I don't see what this change even is.

PR #312, opened 2024-03-24
Refactor associations between items and crafting/tempering materials

danascheider:
## TODO

* [x] Update docs
* [x] Add ADR
* [x] Identify and add missing tests
* [x] Update syncers and Rake task (will be in a separate PR to be merged into this one)

## Context

We are changing our database schema a bit: crafting and tempering materials will now all be coming from the `canonical_materials` table instead of having two separate join models. The `Canonical::Material` model is now a polymorphic join model connecting craftable _or_ temperable items with a material that can be a `Canonical::Weapon`, `Canonical::Ingredient`, or `Canonical::RawMaterial`.

There have been some changes in our taxonomy as well. Instead of having the old `Canonical::Material` class, which could include items that belonged to another class as well (i.e., `Canonical::Weapon` or `Canonical::Ingredient`), we have separated out "pure" materials - those that are exclusively used for building or smithing - into the `Canonical::RawMaterial` model. The `Canonical::Material` class has a polymorphic `:source_material` association that can reference a `Canonical::Weapon`, `Canonical::Ingredient`, or `Canonical::RawMaterial` model. Its mutually exclusive `craftable` and `temperable` associations are used to identify both what item the material can be used for as well as whether it is used for tempering or crafting that item.

This change necessitates significant changes to the code base. I will also need to update JSON data and syncers to reflect the changes. I will do that in a second PR based off this one, in order to keep the PRs as small and focussed as possible. However, since these changes will need to be deployed together, I'll be merging the syncer changes into this PR prior to pushing this one to `main`.

## Changes

* Create `Canonical::RawMaterial` model
* Remove `Canonical::CraftablesCraftingMaterial` and `Canonical::TemperablesTemperingMaterial` models and drop tables
* Update `Canonical::Material` model to  replace the removed models.

### Required Changes

* [ ] Added and updated RSpec tests as appropriate

## Considerations

TODO


spec/models/canonical/armor_spec.rb
```diff
@@ -189,29 +189,44 @@
...
       before do
-        armor.canonical_craftables_crafting_materials.create!(material:, quantity: 4)
+        armor.canonical_crafting_materials.create!(
+          quantity: 4,
+          source_material: create(:canonical_raw_material),
+        )
+        armor.reload
       end
 
       it 'gives the quantity needed' do
```
danascheider, line 203: We need to update this spec since it no longer tests anything meaningful. It should test that the crafting materials for the in-game armour item should be the same as for the corresponding canonical model.

spec/models/canonical/armor_spec.rb
```diff
@@ -189,29 +189,44 @@
...
+        create(
+          :canonical_material,
+          temperable: armor,
+          quantity: 1,
+        )
+
+        armor.reload
       end
 
       it 'gives the quantity needed' do
```
danascheider, line 227: As above so below 😅 This test also needs to be updated to test something meaningful.

spec/models/canonical/jewelry_item_spec.rb
```diff
@@ -150,15 +150,33 @@
...
       before do
-        item.canonical_craftables_crafting_materials.create!(material:, quantity: 2)
+        item.reload
       end
 
-      it 'gives the quantity needed' do
-        expect(item.crafting_materials.first.quantity_needed).to eq 2
+      it 'returns the canonical materials used to craft the item' do
+        raw_material1 = canonical_materials[0].source_material
+        raw_material2 = canonical_materials[1].source_material
```
danascheider, line 177: We can eliminate one of these lines by doing a `map`.

spec/models/weapon_spec.rb
```diff
@@ -717,7 +729,7 @@
           )
         end
 
-        it 'returns an empty array' do
+        it 'returns nil' do
```
danascheider, line 732: Just changed the name of the example to match the expectation.

spec/models/canonical/armor_spec.rb
danascheider: My b, this IS the canonical model. So these tests just really aren't needed anymore.

PR #315, opened 2024-03-27
Update syncers to match new model schema

danascheider:
## Context

In the base branch PR (#312), we have refactored some models, most notably replacing the former `Canonical::Material` with `Canonical::RawMaterial` and making `Canonical::Material` a join model replacing both `Canonical::CraftablesCraftingMaterial` and `Canonical::TemperablesTemperingMaterial`. This change enables us to associate craftable or temperable items with the materials used to craft or improve them, even if those items cannot be reasonably classified as `Canonical::RawMaterial`s.

This PR updates the JSON data, syncers and Rake tasks to work with the new schema.

## Changes

* Reformat JSON data for crafting and tempering materials as well as `Canonical::RawMaterial`s
* Add syncers for crafting and tempering materials
* Update Rake tasks
* Update docs
* Add tests for new syncers

### Required Changes

## Considerations

There is a lot of duplication in the code for the two new syncers. However, there were a few reasons I decided to stick with the two almost identical `Canonical::Sync::CraftingMaterials` and `Canonical::Sync::TemperingMaterials` classes. First of all, since we are removing syncers in an [upcoming epic](https://trello.com/c/LeDebqLf/214-admin-dashboard) and making the database the source of truth, it didn't make sense to spend too much time optimising to reduce tech debt. Similarly, since we won't be adding more syncers like these, there was limited benefit to extracting shared logic into a parent class or mixin, nor would it make sense to attempt modify an existing syncer to provide the functionality needed to its child classes (if that's even possible).


app/models/canonical/sync/association_syncer.rb
```diff
@@ -4,7 +4,7 @@ module Canonical
   module Sync
     class AssociationSyncer < Syncer
       def perform
-        raise Canonical::Sync::PrerequisiteNotMetError.new(prerequisite_error_message) unless prerequisite_conditions_met?
+        raise PrerequisiteNotMetError.new(prerequisite_error_message) unless prerequisite_conditions_met?
```
danascheider, line 7: We don't really need the fully qualified name here, especially since the line is already on the long side.

spec/models/canonical/material_spec.rb
```diff
@@ -86,6 +86,30 @@
     end
   end
 
+  describe 'scopes' do
+    describe '::with_craftable' do
+      subject(:with_craftable) { described_class.with_craftable }
+
+      let!(:included_models) { create_list(:canonical_material, 2, :with_craftable) }
+      let!(:excluded_model) { create(:canonical_material, :with_temperable) }
```
danascheider, line 94: This should go in a `before` block instead of using a memoised helper, since it isn't invoked in the spec.

